PR #2, opened 2026-09-05
Read every Claude Code profile, not only ~/.claude

mhmzdev:
Hey man,
Thank you for open sourcing this. I'm just opening a PR for a problem I faced :)

## What

Anyone who keeps a second Claude Code login apart with `CLAUDE_CONFIG_DIR=~/.claude-work claude` currently gets nothing from Codenotch for that account: its OAuth token is filed under a different keychain service, and its sessions are written to `~/.claude-work/sessions`, so neither the usage ring nor the activity arc ever saw it.

This makes profiles a first-class thing. Every `~/.claude-<slug>` directory Claude Code has run against becomes a **Claude (slug)** ring beside the default one, with its own limits, its own sessions, and its own row in Settings.

## How

- **`ClaudeProfile`** (new) owns the convention: slug from the directory name, provider id `claude-<slug>`, the `sessions` directory, and the keychain service name. Claude Code files a non-default profile's token under `Claude Code-credentials-<first 8 hex of sha256(config dir path)>`; the type derives it the same way (verified against a real keychain).
- **Discovery** runs once at launch and only counts directories carrying a file Claude Code writes on first run (`sessions`, `projects`, `settings.json`, `history.jsonl`, `.claude.json`). An empty or hand-made folder does not become a permanent "sign in" ring. Order is default first, then slugs alphabetically, so rings never swap places between launches.
- **`ClaudeKeychain`** replaces the static credential cache: one reader per profile, so caches and prompts do not bleed between accounts. `ClaudeCredentials.load()` / `forgetCached()` still work and read the default.
- **`ClaudeOAuthProvider`** takes a profile. Rate-limit back-off is stored per provider id; the default keeps its old key.
- **`AppDelegate`** builds one provider and one `ClaudeSessionMonitor` per profile, keyed by the same id, so a work session spins the work ring and nobody else's.
- `ProviderSnapshot.authPrompt` and `ProviderSummary.usesKeychain` recognise `claude-*` ids.

The default profile keeps its id, keychain service and archive keys, so nothing an existing install remembers is lost.

## UI

No change to the notch geometry. Each profile is one more cell of the same shape. What tells them apart:

- Tooltip title: **Claude (work) Usage**, listing only that profile's sessions.
- Settings row: **Claude (work)** — "Max · via Claude Code in ~/.claude-work", with its own switch.
- Sign-in guidance names the exact command, since plain `claude` signs in the default profile only.

## Tests

- New `ClaudeProfileTests` (15): identity, keychain hash, strict directory-name parsing, discovery order, marker rules, per-profile back-off, and two profiles producing two cells.
- Existing keychain and status-message tests extended with a `claude-work` id.

Verified on a Mac with three profiles (`~/.claude`, `~/.claude-work`, `~/.claude-system`): all three discovered, each keychain item read under its derived service name, personal ring showing a live reading, the other two honestly reporting an expired token until Claude Code is next run against them.

Not included: a What's New entry, since that is tied to a version bump.

## Look'n Feel

Settings:
image

Multi-accounts:
image


HoracioEspinosa:
This is a very useful implementation!, im just implementing that in my fork, thanks!

HoracioEspinosa (approved):

vinzdg:
Merged in 1b5b63e (shows closed rather than merged — history quirk on my end, not a rejection). This conflicted with #6 in ClaudeOAuthProvider.swift since both rewrote the token path; I combined them by hand so the per-profile ClaudeKeychain design carries #6's back-off fix and testing seam, and both test suites pass together. Thanks for the multi-profile support.

vinzdg:

mhmzdev:
> Update: this shipped in **v1.4.0**, released just now.

So cool! Thank you so very much ✌️

PR #4, opened 2026-09-05
Stop Cursor chats reporting work that already ended

ppuzio:
## Cursor chats stay "working" for ever

`composerHeaders` never clears `unfinishedRunAt` when a run doesn't end cleanly — editor killed mid-run, or the run abandoned with the editor still open. Reading the field's presence as "busy" leaves the row spinning indefinitely; mine showed three chats at 28+ hours.

It can't date the run either. Measured against a real store it equals `createdAt` on every row that carries it, and a row with two turns seven minutes apart still reports the first. It's a flag.

So busy now means `conversationCheckpointLastUpdatedAt` — which Cursor moves on every message and tool result — is after the editor's launch and within `staleAfter`. Falls back to `lastUpdatedAt`, which a quarter of rows carry instead. `staleAfter` is 15 min, injected like the Codex and Antigravity monitors: the longest gap between writes inside a run, over 33,484 of them, was 826s, so 10 min would have cut live runs.

`NSRunningApplication.launchDate` is optional and usually absent (105 of 123 running apps here, Finder and Chrome included), so "running, start unknown" answers `.distantPast` rather than collapsing into "not running". Bundle-id lookup falls back to the bundle name for the same reason — a miss would silently retire every Cursor row.

Also fixes `DateFormatter` ignoring an injected calendar's time zone, in `ResetCopy` and again in `UsageBlock.summary`.

14 tests.


vinzdg:
Merged in 1b5b63e — GitHub shows this as closed rather than merged because it landed via a manual merge across a history rewrite on my end, not because anything was rejected. All 6 new tests pass in the combined tree. Thanks for the detailed writeup and the real percentile data behind `staleAfter`.

vinzdg:

PR #6, opened 2026-09-06
fix(claude): recover usage readings without an app restart

oasis1992:
# Claude usage stopped being read, with no way to recover it

Two defects, one behind the other.

### 1. The notch stopped showing usage and never recovered

Once it failed, it did not come back. Not on the next tick, not an hour later, not
when the cause of the failure was long gone. The only way out was restarting the
app.

Meanwhile:

- Claude Code kept working normally
- the session was active and the token sat in the keychain, readable
- the Settings window reported the `Max` plan without trouble
- the notch, in the same process, said **"Sign in to Claude Code to read your usage"**
- and listed the running Claude Code sessions directly underneath that message

A real case: 45 minutes straight asking for a sign-in, with the panel itself
showing `codenotch-03 · working · just now` two lines below.

### 2. "Allow access…" was no way out of it

It was the only remedy the interface offered, and it was dead: pressing it did
nothing at all, not even raising the macOS dialog. It was also permanently on
screen, including when access was granted and there was nothing to repair.

Both defects share a cause: the same lock that froze the reading was intercepting
the button before it could reach the keychain.

### After this PR

- the reading recovers on its own on the next tick — 60 seconds while you are working
- the button works, and appears only while macOS is actually refusing

## Screenshots

**Before — the remedy offered next to a working account**

The Allow access button shown next to a working account

> Access granted, readings fine, and the row still offering to ask macOS again.

**Before — the wrong guidance after a declined prompt**

The row telling a signed-in user to run Claude Code after a declined prompt

> The user is signed in and Claude Code is running. The prompt was declined; that
> is what needs fixing, and it is not what the row says.

**After — nothing on offer when there is nothing to fix**

The same row after the fix, with no remedy on offer

> Same account, same working readings, and the row no longer offering a remedy.

## What was wrong

`ClaudeOAuthProvider` kept its own auth back-off on top of the one
`CredentialCache` already implements:

```swift
// currentToken() — bailed out before ever reaching the keychain
if let lastAuthFailure, Date().timeIntervalSince(lastAuthFailure) < authRetryDelay {
    throw UsageProviderError.needsAuth
}

// fetchSnapshot() — re-stamped the window with the failure it had just caused
} catch UsageProviderError.needsAuth {
    lastAuthFailure = Date()
    throw UsageProviderError.needsAuth
}
```

The poll runs every 60 seconds and the window was 5 minutes, so every tick
re-stamped it before it could expire. A single `needsAuth` locked the provider for
the life of the process.

Production logs show 45 minutes of `claude failed: needsAuth` without **one**
`keychain read failed` line. That message is emitted for any non-success
`OSStatus`, so its absence proves `read()` was never called: the keychain went
unconsulted for three quarters of an hour.

`account()` does not go through that path — it calls `ClaudeCredentials.load()`
directly — which is why Settings kept showing the plan while the notch insisted
there was no session. Same process, two paths, one lock between them.

### What armed it

```
16:07:36.103  keychain read failed: OSStatus -25320 (In dark wake, no UI possible)
16:07:36.104  claude failed: needsAuth
```

`errSecInDarkWake`. The Mac was in dark wake — awake for background work, with the
display off and therefore no way to show a dialog. The keychain did not deny
access; it said *not now*.

`ClaudeCredentials.wasRefused()` recognises only three statuses:

```swift
status == errSecAuthFailed
    || status == errSecUserCanceled
    || status == errSecInteractionNotAllowed
```

`-25320` is not among them, so a transient environmental condition landed in
`needsAuth` — the one error that arms the lock. The machine woke up, Claude Code
carried on working, and the notch stayed stuck for the next 45 minutes.

A detail that stings: `UsageStore.start()` registers an observer on
`NSWorkspace.didWakeNotification` that calls `refreshNow()`, precisely because
*"waking up is the one moment the numbers are guaranteed to be wrong"*. That safety
net fired and the lock swallowed it like every other tick.

## How it was reproduced

A dark-wake read cannot be summoned on demand, and neither can a 401 from the real
endpoint. But both end in the same `needsAuth`, and a 401 **can** be manufactured
by pointing the provider at something we control.

A local server on `127.0.0.1:8899` answered 401 or 200 from a switch file, with the
endpoint temporarily redirected to it. Two runs per side, ad-hoc signed, with an
active Claude Code session keeping the poll at its 60-second rate.

| | before | after |
|---|---|---|
| requests while the endpoint returns 401 | 2, then never tries again | 2 per tick, keeps trying |
| requests once it returns 200 | 0 | 1, on the next tick |
| time to recover | never (capped at 241s) | 60s |

```
before    flip to 200 at 17:47:04 → still failing at 17:50:18, count frozen at 3
after     flip to 200 at 17:52:33 → 200 served at 17:53:32
```

60 seconds is exactly `UsageStore.refreshInterval`: the fix adds no faster path, it
lets the tick that already existed reach the end.

## The fix

It is a deletion. `CredentialCache` already does this back-off correctly one layer
down:

- its suppressed-retry branch returns **before** the `do/catch`, so it never
  re-stamps itself
- it keys on the keychain item's modification date rather than a clock, so a token
  Claude Code has just rotated is picked up on the next tick
- `forget()` clears it outright, which is what makes "Allow access…" work

Removed `lastAuthFailure`, `authRetryDelay`, the guard in `currentToken()` and both
assignments in `fetchSnapshot()`. The three real intervals are untouched:
`UsageStore.refreshInterval` (60s), `UsageStore.idleRefreshInterval` (5min) and
`CredentialCache.retryAfterFailure` (5min).

## The Allow access button

Two problems, both in the same row.

It was **dead** after a `needsAuth`. `reauthorize` drops the in-memory credential
and refreshes, but the refresh hit the lock before reaching the keychain, so
pressing it did nothing whatsoever. Removing the lock fixes that.

It was also **always on screen**, for any keychain-backed provider, whether or not
there was anything to repair. A remedy offered next to a working account is
indistinguishable from a broken button — which is exactly how this was reported.

Gating it turned out to be less obvious than it looks. The snapshot's status is not
the place to read a refusal from: `supersedesHistory` deliberately keeps the
previous reading **and its status** through an `accessDenied`, because the number is
still true and discarding it would punish someone for pressing the wrong button. The
refusal never reaches the row.

They are two different questions — "how good are the numbers I am showing" versus
"did macOS let us in last time" — so the store now remembers the second separately
in `refusedAccess`, cleared as soon as a fetch succeeds. The snapshot is untouched
and the notch keeps showing dimmed readings exactly as before.

The row's guidance follows the same fact: it no longer tells a signed-in user to run
Claude Code when the real problem is a declined keychain prompt, which is the case
`UsageProviderError.accessDenied`'s own doc comment warns against.

## Tests

`fetchSnapshot()` had no coverage at all — only the pure helpers were tested, which
is how this reached production. Added a `URLProtocol` stub (the repo had none) and
injected the credential read.

Every assertion is about **whether the provider asked again**, because the broken
version returned the correct error while touching nothing:

- a 401 does not stop the next tick from reaching the endpoint
- a failed keychain read does not stop the next tick from reading again
- three ticks against a rejecting endpoint produce six attempts, not two

All three fail against the previous implementation.

For the row, five more — including the one that matters most: a refusal arriving
when a good reading is already in hand. The first attempt at this gating passed its
tests and did nothing on screen, because the fake provider had no archived reading
and so never entered the branch that preserves history.

Unrelated to this change: `ResetCopyTests` has 3 pre-existing failures from the
machine's locale rendering "10:13 p.m." with full stops. Verified on a clean tree.

## Not in this PR

- `-25320` should not be classified as "no session". With this fix it costs one tick
  instead of the whole session, but it still tells a signed-in user to sign in.
- A persistent 401 re-reads the keychain on every tick, which under a plain "Allow"
  rather than "Always Allow" means two dialogs a minute. It wants the shape
  `retryNoEarlierThan` already has for 429s.


nixonsucuc:
Facing the same issue myself. Hope this one gets merged soon!!

vinzdg:
Merged in 1b5b63e — same note as elsewhere: shows as closed rather than merged due to a history quirk on my end, not a rejection. This one needed reconciling with #2's ClaudeKeychain refactor since both touched ClaudeOAuthProvider's token path; combined so the multi-profile design keeps your loadCredentials injection seam and all your tests. Thanks for tracking down the dark-wake OSStatus.

vinzdg:

PR #8, opened 2026-09-06
Build without the maintainer's certificate

MintarasGrinius:
Debug build/test ad-hoc sign (matches CONTRIBUTING.md); DEVELOPER_DIR only defaults when Xcode exists. Release archive keeps Developer ID.

Test: make build + make test green (563 tests).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,21 @@
+# Default to standard Xcode location only if it exists and the user hasn't
+# already chosen a toolchain via `sudo xcode-select -s` / $DEVELOPER_DIR.
+# Unconditionally exporting a path that doesn't exist breaks `make` on machines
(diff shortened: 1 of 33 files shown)


vinzdg:
Landed in v1.5.0, refined: your fix would have also ad-hoc signed the maintainer's own local `make run`, and an ad-hoc identity changes on every rebuild — which throws away the whole reason `project.yml` pins a stable Developer ID (so a keychain "Always Allow" grant survives a rebuild instead of re-prompting every time). The merged version only falls back to ad-hoc signing when `security find-identity` doesn't find that certificate at all, so it's invisible on the maintainer's machine and fixes it for everyone else. Thanks for catching this — it's been a real gap since the repo went public.

PR #9, opened 2026-09-06
Codex: skip empty premium snapshots when reading rollouts

MintarasGrinius:
Merge after #8 (stacked; diff shrinks once #8 lands — this PR's own commit is the Codex one).

Newer Codex builds write a premium-typed snapshot with null windows into the newest thread; the provider read that thread only and reported 'no usage windows', hiding older real readings. Now it walks recent rollouts newest-first and takes the first with usable windows (its own timestamp decides staleness). Activity monitor unchanged.

Test: make test green, incl. new CodexRolloutFallbackTests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,21 @@
+# Default to standard Xcode location only if it exists and the user hasn't
+# already chosen a toolchain via `sudo xcode-select -s` / $DEVELOPER_DIR.
+# Unconditionally exporting a path that doesn't exist breaks `make` on machines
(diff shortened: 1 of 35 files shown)


vinzdg:
Thanks for this — landing #14 in the same batch replaced the rollout-based reading entirely with a live call to Codex's own account endpoint, so the rollout-parsing path these fixes improved doesn't exist anymore. Nothing wrong with the fix itself, it's just superseded by the redesign. Appreciate the work regardless.

PR #10, opened 2026-09-06
Add Grok SuperGrok weekly pool via signed-in WebView

MintarasGrinius:
Merge after #8, #9 (stacked; own commit is the Grok one).

New provider: weekly SuperGrok pool from grok.com's own Usage-card RPC (GetGrokCreditsConfig, gRPC-web/protobuf, no published schema — layout ported from CodexBar/alex-core and pinned by fixtures). Sign in once via Right-click -> Sign in to Grok; free-tier accounts have no weekly pool and will read as an error until that shape is captured.

Test: make test green, incl. new GrokUsageTests (6 tests).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,21 @@
+# Default to standard Xcode location only if it exists and the user hasn't
+# already chosen a toolchain via `sudo xcode-select -s` / $DEVELOPER_DIR.
+# Unconditionally exporting a path that doesn't exist breaks `make` on machines
+# with only the Command Line Tools installed (xcrun: missing DEVELOPER_DIR).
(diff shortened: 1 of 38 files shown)


vinzdg:
Thank you for this. #12 and #15 both independently arrived at reading `~/.grok/auth.json` directly (the CLI's own stored session, token-authenticated REST call) rather than a signed-in browser session — and I confirmed that approach gets the same weekly Grok Build credits without needing a browser at all, which fits how every other provider in this app already works. Went with #15's version since it was the most complete. Appreciate the investigation into the gRPC-web RPC regardless — genuinely useful if the local-file approach ever stops working for some account shape.

PR #11, opened 2026-09-06
Add OpenCode Go plan windows via official usage endpoint

MintarasGrinius:
Merge after #8, #9, #10 (stacked; own commit is the OpenCode one).

New provider: Go plan rolling/weekly/monthly windows from the official endpoint, borrowing the opencode-go key — no second sign-in. Verified live (HTTP 200). Zen pay-as-you-go credit balance has no API upstream, so Go windows only.

Test: make test green (563 tests), incl. new OpenCodeUsageTests + OpenCodeCredentialsTests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,21 @@
+# Default to standard Xcode location only if it exists and the user hasn't
+# already chosen a toolchain via `sudo xcode-select -s` / $DEVELOPER_DIR.
+# Unconditionally exporting a path that doesn't exist breaks `make` on machines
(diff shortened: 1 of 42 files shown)


vinzdg:
Landed in v1.5.0 — the OpenCode provider, credentials reader, usage parser, and tests came in essentially as you wrote them. It shows here as still open rather than merged because it had to be applied on top of a different Grok/Codex base than the one it was stacked on (see #10, #14 for why), so the commits don't share the exact SHAs GitHub checks for — the code is the same, just re-applied. Thanks for this.

PR #12, opened 2026-09-06
Feat/fable and grok usage

maheshauti96:


vinzdg:
Thanks for this. Went with #15's Grok implementation instead since it was more complete (an activity monitor, fuller test coverage) — your GrokCredentials.swift independently arrived at the same ~/.grok/auth.json approach as #15, which was useful confirmation that it's the right mechanism.

One thing worth flagging for next time: both #12 and #15 hardcoded a specific day of the month as the account's subscription renewal date (17th for Claude, 7th for Codex, 18th for Grok) and applied it to every user. I couldn't find where those numbers come from any API response — they read as one personal account's own billing dates presented as universal facts, which would show the wrong renewal date to almost everyone else. I dropped that part of #15 for the same reason. If a real per-account charge date becomes available from one of these endpoints, that'd be a great follow-up.

PR #13, opened 2026-09-06
fix: stop disconnected providers reading accounts or restoring usage

canberkys:
Turning a provider off could still read its account when Settings opened. If a refresh was already running, its late response could also restore the removed ring and re-save the forgotten usage reading.

This change makes the disconnected state apply to account summaries and to every stage of a usage refresh:

- Keep disconnected providers listed in Settings without calling `account()`.
- Invalidate outstanding responses when the connection changes or sign-out begins, including a quick off/on toggle.
- Skip providers disconnected while queued behind another request; discard late successes and failures before updating the archive or access-refusal state.
- Recheck accumulated results before publishing a full refresh, and clear the single-provider spinner when a result is discarded.

Already-sent requests may finish; their results are discarded. The provider endpoints and polling intervals are unchanged.

Validation:

- Added 9 regression tests using controlled fake providers and isolated UserDefaults suites, without live credentials or network requests.
- Ran the first 8 regression cases against unchanged v1.4.0 (`60bafc2`): all 8 failed, reproducing the defects. All 9 pass with the fix.
- Full suite: 555 tests passed on Apple silicon, macOS 26.6.2.
- `git diff --check` passed.

The full suite was run with the existing XcodeGen scheme and local signing overrides because the current project requires the maintainer's certificate (already addressed separately by #8):

```sh
xcodegen generate
xcodebuild -project Codenotch.xcodeproj -scheme Codenotch \
  -destination 'platform=macOS,arch=arm64' -configuration Debug \
  -derivedDataPath build/DerivedData CODE_SIGN_IDENTITY=- DEVELOPMENT_TEAM= test
```

### DIFF
--- a/Sources/Model/UsageStore.swift
+++ b/Sources/Model/UsageStore.swift
@@ -14,13 +14,20 @@ final class UsageStore: ObservableObject {
     @Published private(set) var refusedAccess: Set<String> = []
 
     private let providers: [UsageProvider]
+    /// A response belongs to the connection that started it. Checking only
+    /// `disconnected` would accept an old response after a quick off/on toggle.
+    private var connectionVersions: [String: UUID] = [:]
     /// Providers the user has switched off. They are not fetched at all — their
     /// credential is never read, which is the whole point of switching one off.
     /// Filtering the results afterwards would still touch the keychain.
     @Published var disconnected: Set<String> = [] {
         didSet {
             guard disconnected != oldValue else { return }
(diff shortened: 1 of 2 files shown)


vinzdg:
Merged as-is in v1.5.0 — real bug, confirmed independently (providerSummaries was calling account() unconditionally, even for switched-off providers), well tested. Thanks for the thorough repro against v1.4.0.

PR #14, opened 2026-09-06
fix: read Codex account limits from the usage endpoint

TrapstarKS:
Codex usage came from the local app server with a rollout-log fallback. Those readings could be stale or describe a different allowance from the account's main 5-hour and weekly limits, leaving the notch showing the wrong quota.

This change reads the current account limits from https://chatgpt.com/backend-api/wham/usage:

- Show the main 5-hour and weekly windows when present, identifying each by its duration.
- Exclude Spark and other additional quotas. Accounts with only a weekly limit show only that ring.
- Remove the old app-server bridge and rollout usage reader, and discard cached readings with the old window identifiers.
- Read the existing Codex credentials on each request so token rotation is picked up, without refreshing or changing them.
- Respect `Retry-After` on HTTP 429, with a minimum 60-second delay persisted across app restarts.

Validation:

- Updated parser coverage with 3 focused tests for both main windows, a weekly-only response, and reset countdowns, including exclusion of additional quotas.
- Full suite: 523 tests passed on Apple silicon.
- A live request returned 5% weekly usage and no 5-hour window; the provider produced a weekly-only snapshot.
- `git diff --check` passed.

The full suite was run with local signing overrides because the project requires the maintainer's certificate:

```sh
xcodegen generate
xcodebuild -project Codenotch.xcodeproj -scheme Codenotch \
  -destination 'platform=macOS,arch=arm64' -configuration Debug \
  CODE_SIGN_IDENTITY=- DEVELOPMENT_TEAM= CODE_SIGN_STYLE=Manual test
```


vinzdg:
Merged in v1.5.0, with one fix: the parser only recognised exactly 18,000s (5h) and 604,800s (7d) and silently dropped anything else. I tested the live endpoint against a real free-plan account and its actual primary_window was 2,592,000s (30 days) — a real, active limit that would have been discarded entirely, leaving that account's ring reporting nothing metered. The label is now derived from whatever duration the account actually reports (reusing the same generic logic the old rollout parser had), so a monthly window — or whatever Codex introduces next — gets a sensible label instead of silently vanishing. Thanks for the endpoint switch itself; the numbers otherwise matched exactly.

PR #15, opened 2026-09-06
Codenotch 1.5.0: Grok ring and charge dates

pgm-norbert:
## Summary

Grok is a fifth provider on the notch. SuperGrok weekly Grok Build credits are read from the CLI session in `~/.grok/auth.json`. The tooltip names the **charge day** — Grok the 18th, Claude the 17th, Codex the 7th — not a usage-window reset (Grok `/billing` ending on the 1st is the ledger, not the bill).

## Test plan

- [x] `make test` — 581 tests, 0 failures (ad-hoc local signing; Developer ID left in `project.yml`)
- [ ] Hover Grok / Claude / Codex — first line under the title is Renews Sep 18 / Sep 17 / Sep 7
- [ ] Settings account rows show the same copy
- [ ] Grok ring is weekly credits, not a dash

Cannot attach a notarized dmg from this machine (no Developer ID `6WFPL8B9FB` / no `UsageNotch` notary profile). Other Macs can clone this branch and `make run` until the maintainer cuts the Sparkle build.


vinzdg:
The Grok provider landed in v1.5.0 — thanks, it was the most complete of three independent Grok submissions this batch, and the activity monitor and test coverage made it the right base.

I dropped the charge-date feature entirely, for all three providers it touched. BillingAnniversary hardcoded specific days of the month (Grok 18th, Claude 17th, Codex 7th) as universal constants, applied to every account regardless of when that account's subscription actually renews — I couldn't find where those numbers come from any real API response, and they read as one account's own billing dates presented as fact for everyone else. That's the one thing this app has been careful never to do anywhere else (see the Fidelity types throughout — a derived number always gets a "~", nothing is ever shown as official that wasn't). If a real per-account charge date turns up in one of these responses, happy to revisit.

PR #19, opened 2026-09-07
Read Cursor usage from cursor-agent when the editor is missing

tscuite:
## Summary

Codenotch already borrows the Cursor editor's SQLite session. That leaves CLI-only machines — `cursor-agent login`, no Cursor.app — with a blank ring and no way to sign in, because there is no editor to open and Codenotch never had its own Cursor login.

The CLI session is the same cookie the editor uses. `/api/usage-summary` wants `WorkosCursorSessionToken={accountID}::{accessToken}`. `cursor-agent` files the JWT under keychain service `cursor-access-token` and the account id in `~/.cursor/cli-config.json` (`authInfo.authId`, then numeric `userId`, then the JWT `sub`). A token-only cookie or a Bearer header is 401.

This change:

- Keeps the editor store first, so a laptop with both tools cannot flip accounts.
- Falls through to the CLI session only when the editor is missing or signed out.
- Caches the keychain read (`CredentialCache`) and surfaces `accessDenied` / `credentialExpired` the same way Claude Code does.
- Marks Cursor as keychain-backed so a declined prompt can offer "Allow access…".
- Opens Cursor when the editor is installed; otherwise names `cursor-agent login` instead of an "Open Cursor" button that would do nothing.
- Leads the Cursor ring with **Auto usage** (`autoPercentUsed`). `totalPercentUsed` is a blend of Auto and API and is not a dashboard row, so a 20% Auto / 100% API account no longer reads as 25%.

## Test plan

- [x] New `CursorCredentialsTests` cover cookie construction, editor-over-agent precedence, agent fallback, `authId` / `userId` / JWT `sub`, expired JWT → `credentialExpired`, and the two sign-in routes. The agent token is injected so tests never touch the live keychain.
- [x] `CursorUsageTests` cover the Auto headline, omitting the blended total, and the 20% vs 25% case.
- [ ] `make test` — could not run here: this machine has Command Line Tools only, no Xcode.app, so `xcodebuild` refuses. Please run `make test` on a Mac with Xcode.
- [ ] Mac with only `cursor-agent login` (no Cursor.app): Cursor ring appears after Allow / Always Allow; Settings shows the CLI email via `cursor-agent`.
- [ ] Mac with the editor signed in: behaviour unchanged; keychain is not read.
- [ ] Mac with the editor installed but signed out, CLI signed in: ring comes from cursor-agent.
- [ ] Decline the keychain prompt: "Allow access…" comes back; Always Allow stays quiet.
- [ ] Cursor ring matches Plan & Usage **Auto**, not the blended total. Tooltip has Auto and API only.

No tokens, cookies, emails, or live account ids are in this PR. Test cookies use placeholder ids (`auth0|user_CLI`, `cli@example.com`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 build/
 .build/
 dist/
+Package.resolved
 DerivedData/
 *.xcodeproj
 *.xcworkspace
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 .DS_Store
 build/
+.build/
+dist/
 DerivedData/
 *.xcodeproj
 *.xcworkspace
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ two never disagree.
 | Provider | Source | How |
 |---|---|---|
 | **Claude Code** | official | The OAuth token in the login keychain, against the same endpoint Claude Code's own `/usage` uses. |
-| **Cursor** | official | The editor's own signed-in session, read from its local SQLite state — no separate sign-in. |
+| **Cursor** | official | The editor's signed-in session in its local SQLite state, or the `cursor-agent` login in the keychain — no separate sign-in. |
 | **Codex** | official | ChatGPT's usage endpoint, using the local Codex sign-in. Shows the 5-hour and weekly limits when available. |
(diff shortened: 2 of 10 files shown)


vinzdg:

A CLI-only Cursor login now gets a ring — Codenotch reads `cursor-agent`'s session when the editor isn't installed.

Genuinely appreciated.

PR #20, opened 2026-09-07
Let the user order the provider rings

oasis1992:
`TASKS.md` has carried `- [ ] Drag-to-reorder providers` under M5 since the
milestone was written. This implements it.

The order the notch draws was a hardcoded array literal in `AppDelegate`, so
someone who lives in Codex reads past three rings they do not care about every
time they glance at the strip.

## What changed

Settings splits **Integrations** into **Connected** and **Not connected**, and a
connected row can be dragged by its handle. Ordering only means something for the
first group — a provider switched off draws no ring, so dragging it was arranging
something that is not on screen.

The notch, rings in the chosen order

*The strip, drawn in the order set below.*

Settings — Connected

*Connected rows carry a handle. Account addresses redacted for the screenshot.*

Settings — Not connected

*Switched-off providers have no handle — nothing to place.*

## Design decisions worth reviewing

**Stored as ids, not as a `sortIndex`.** This diverges from the design spec
deliberately. The provider set is not fixed — Claude Code contributes one per
`~/.claude-<slug>` found at launch — so an index is a number about a list that no
longer exists. `ProviderOrder` reconciles a stored order against the live
registry in both directions: an id the order has never seen is **appended rather
than dropped** (a provider added by a new version must not silently vanish), and
a stored id whose provider is gone **keeps its place rather than being pruned**
(a `~/.claude-work` that is not on this Mac today). The archive already carries a
`perplexity` entry from a provider that no longer exists, so ids outliving their
provider is a shape this repo already has.

**Applied in `UsageStore`, not at each consumer.** There are two consumers — the
notch reads `snapshots`, Settings reads `providerSummaries` — and they have to
agree by construction rather than by two call sites remembering to.

**Reordering never refetches.** Dragging a row is a question about layout;
answering it by re-reading every credential would spend Claude's rate-limit
budget on nothing. `order.didSet` re-sorts what the store already holds. Pinned
by `testReorderingMovesTheRingsWithoutRefetching`.

**A provider switched back on joins the end of the connected list** rather than
reclaiming an older position. Restoring its old place would make off/on a perfect
round trip, which is tempting — but it lets a row that was not on screen outrank
one the user deliberately dragged to the top while it was away.

**An empty stored order means "never chosen"**, kept distinct from "chose the
current default", so an existing install updates into the order it already had
and nobody wakes up to moved rings.

## Implementation notes

`ForEach.onMove` does nothing inside a `Form` on macOS — it is honoured by
`List`, and on iOS a `Form` *is* a `List` underneath, which is why it compiles
and looks like it should work. Converting the sheet to a `List` to get it would
restyle all four sections, so the drag uses `.onDrag`/`.dropDestination`, which
are plain view modifiers.

The rows rearrange **during** the drag rather than settling once on release.
`dropDestination` does not hand over its payload until the drop, so the dragged
id is captured at drag start and held in a plain reference (`DragState`) that no
row observes — as SwiftUI state it re-rendered every row twice per gesture.

`GrabCursor` is the one piece that looks like overkill and is not. AppKit resets
the cursor when a drag session ends and does not re-evaluate until the mouse next
moves, so releasing the button and holding still left an arrow on a grip that was
perfectly grabbable. `.pointerStyle` rides the same mechanism and inherits the
limitation; `invalidateCursorRects(for:)` is the only lever that forces a
re-evaluation under a stationary pointer. It must sit *over* the content —
behind it, SwiftUI's own pointer regions win and the rect is never consulted.

## Tests

19 new tests, placed in the existing classes closest to each concern per
CONTRIBUTING: the pure rule in a new `ProviderOrderTests`, store plumbing in
`ProviderDisconnectionTests`, persistence in `PreferencesTests`.

`571 tests, 1 failure` — the failure is `ResetCopyTests.testAbsoluteTimeUsesAColon`,
which is **pre-existing and unrelated**: it asserts the formatted reset time
contains no `.`, and a Spanish-locale machine renders `11:32 p.m.`. It fails on
`main` without this branch. Everything else is green.

## Not included

- The Settings window is still a fixed 560pt, so **Not connected** needs a
  scroll to reach. Left alone rather than changing the window's shape inside a
  feature PR.
- `CODENOTCH_DEMO=1` keeps the fixture order; it bypasses the store entirely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,12 @@ them, and its ring appears. Switching a provider off in Settings stops its
 credential being read at all and forgets the readings taken from it; it does
 not sign you out of the tool that owns the account, and the row says so.
 
+Settings lists the connected providers in the order the notch draws them, and
+you can drag one by its handle to move it. The order is remembered across
+launches. A provider you switch back on joins the end of that list rather than
+reclaiming an older position, so nothing you cannot currently see jumps ahead
+of something you placed deliberately.
+
 It also answers **"is it still working?"** — a thin arc spins inside a
 provider's ring while a session is busy, and becomes a pulsing amber ring when
 one is blocked waiting on you. Hover for every live session by name, where it
--- a/Sources/App/AppDelegate.swift
+++ b/Sources/App/AppDelegate.swift
@@ -72,7 +72,12 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                     + [CursorLocalProvider(), CodexLocalProvider(), AntigravityProvider(),
                        GLMProvider(), GrokLocalProvider(), OpenCodeProvider()]
                     + webProviders,
-                disconnected: preferences.disconnectedProviders
+                disconnected: preferences.disconnectedProviders,
+                // Passed at construction, not left to the sink below, for the
+                // same reason `disconnected` is: the sink delivers a run loop
+                // turn later, so without this every launch draws the built-in
+                // order for a frame and then visibly shuffles.
+                order: preferences.providerOrder
             )
 
             // The stored edge goes in before the panel is ever put up. The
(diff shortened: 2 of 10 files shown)


vinzdg:

Settings now splits into Connected and Not connected, and a connected row drags by its handle to set the order the notch draws the rings in.

Genuinely appreciated.

PR #23, opened 2026-09-07
feat: read usage without the keychain, and attach the release dmg

### LINKED ISSUE #18: Publish a precompiled .dmg in releases
The v1.4.0 release has no assets attached, and building from source requires full Xcode (project.yml targets a Developer ID signing identity and asset catalogs, so Swift Package Manager / Command Line Tools alone can't build it).

Would it be possible to attach the notarized .dmg (from `make release`) to the GitHub release? That would let users install without a ~40GB Xcode download just to try a small menu-bar app.

mulhamna:
## Summary

- Claude Code files a **new** keychain item on every token rotation, and the new item's access list does not carry Codenotch. An "Always Allow" is therefore good until the next rotation and no longer, so the password dialogue came back roughly hourly. Both Claude and Antigravity already expose the same figures through something that needs no keychain access at all — `claude "/usage"` and Antigravity's own language server — so those are asked first, with the existing token paths kept underneath as fallbacks.
- `make build` failed on any machine without the maintainer's Developer ID certificate, contradicting CONTRIBUTING.md. The `DEV_SIGN` guard added in 4123c816 tested `grep -c` against the empty string; `grep -c` always prints a count and reports absence through its exit status, so the condition could never hold.
- Closes #18: `make release` already produced a notarized dmg and copied it into `site/` for the Sparkle feed, but nothing ever put it on the release page, so v1.4.0 and v1.5.0 carry no assets.

## Changes

**Build and release**
- `Makefile`: compare the Developer ID probe against `0` rather than the empty string, so Debug ad-hoc signing actually applies.
- `Makefile`: new `publish` target — `gh release upload $(TAG) $(DMG) --clobber`, with `TAG` defaulting to `v$(MARKETING_VERSION)` read from `project.yml`. Kept out of the `release` chain deliberately: that chain is entirely local and this writes to the remote, so it should be a decision rather than a side effect.

**Claude**
- `Sources/Providers/ClaudeUsageCLI.swift` (new): finds the `claude` binary where its installers put it, runs `/usage` in a scratch directory under a watchdog, and reads the result into `LimitWindow`s. Ids and labels come from the existing `UsageResponse.label(forKind:)` and `displayOrder`, so a reading archived under one source still matches when the other takes over.
- `ClaudeOAuthProvider` asks the CLI first and falls through to the token path on any failure — absent, signed out, wording changed, wedged and killed. A machine without Claude Code behaves exactly as it did.
- Asked ahead of the endpoint's back-off deadline: the two do not share a rate limit, so a 429 on one is no reason to darken a ring the other can still fill. Throttled to one subprocess every five minutes, since the store polls every 60s while a session is busy.
- `account()` answers from Claude Code's own config rather than the token, so opening Settings no longer prompts either. That trades the plan name for the signed-in address — the field the token could never fill, and the only question two Claude rings raise.
- Reset times are read in both spellings Claude Code uses: `2:59pm`, and `3pm` on the hour. Reading only the first left a window with no reset for one hour in sixty.

**Antigravity**
- `AntigravityProvider`: the language server moves from third to first. It already holds the credential and the client identity Google insists on; being asked after a keychain read meant that dismissing the prompt produced an empty ring while the server that would have answered sat running on the same machine, never asked.
- `account()` reads the keychain item's attributes rather than its contents to say an account exists — those are not behind the access prompt the secret is.
- The `:loadCodeAssist` body is no longer decoded on this path. Only its status code was load-bearing, and the local it was decoded into was already unused; the compiler had been saying so.

## Test plan

- [x] `make test` passes with no override and no Developer ID certificate in the keychain — which is also the proof of the signing fix. Before it, `make build` stopped at `No signing certificate "Developer ID Application" found`.
- [x] No new compiler warnings; the four that remain are all present on `main`.
- [x] Claude, end to end, with the app launched the way a user launches it — `env -i open Codenotch.app`, so it inherits launchd's environment and not a developer shell's. `lsof` recorded **no network connection from the app at all** while `ps` recorded `claude /usage` as its child. Since the token path cannot produce a reading without calling `api.anthropic.com`, that is the reading coming from the CLI and the keychain going untouched.
- [x] The figures match: the app read `session 79% / weekly 8%, resetting 15:00 and Sep 14 06:00` against `claude "/usage"` reporting `79% / 8%, resets 3pm / 6am` at the same minute.
- [x] Claude fallback: with the binary moved aside, no subprocess is spawned, nothing is logged as failed, and the ring still fills from the token path.
- [x] Reset dates are pinned across the New Year in both directions, on the hour, at midnight, and against a zone name that contains "am".
- [ ] `make publish` has **not** been run — it needs write access to this repository. Verified only as far as `make -n publish`, which expands to `gh release upload v1.5.0 build/release/Codenotch.dmg --clobber`.
- [ ] Antigravity has **not** been verified end to end. It is not installed on the machine this was written on, so its language server could not be started; that path is covered by unit tests with an injected source rather than by a real reading. The keychain fallback underneath it is unchanged.

One thing worth saying plainly: #18 also mentions needing a full Xcode to build from source. The signing half of that is fixed here — `xcodegen` and `xcodebuild` are still required, and this does not change that.

## Evidance
Screenshot 2026-09-07 at 14 50 35

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,13 @@ DEST    := platform=macOS,arch=arm64
 # changes nothing: project.yml's stable identity is what keeps a keychain
 # "Always Allow" grant alive across rebuilds, and forcing ad-hoc there would
 # throw that away and bring the prompt back on every `make run`.
-ifeq (,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
+#
+# Compared against `0`, not against empty: `grep -c` prints a count on every
+# run and only sets its exit status, so the empty-string test this started as
+# was never true and the ad-hoc fallback never applied. A contributor with no
+# certificate got `No signing certificate "Developer ID Application" found`
+# from `make build` — the very thing this block exists to prevent.
+ifeq (0,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
 DEV_SIGN := CODE_SIGN_IDENTITY="-" DEVELOPMENT_TEAM="" CODE_SIGN_STYLE=Automatic
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ APP_NAME    := Codenotch
 NOTARY_PROFILE := UsageNotch
 DMG := $(RELEASE_DIR)/$(APP_NAME).dmg
 
-.PHONY: archive dmg notarize release verify-release
+.PHONY: archive dmg notarize release verify-release publish
 
 # Release configuration, exported with the Developer ID identity. `xcodebuild
 # archive` + `-exportArchive` rather than a plain build: it re-signs the bundle
@@ -156,6 +156,25 @@ appcast: $(DMG)
 
 release: notarize verify-release appcast
 	@echo "Notarized: $(DMG)"
+
+# The GitHub release page is where someone who has never installed the app
+# looks first; the appcast feed is only ever read by copies already running.
+# The same notarized dmg belongs in both, and until it was in both the release
+# pages carried no assets at all — leaving a full Xcode install as the only way
+# to try the app.
+#
+# Deliberately not part of `release`: every other target here is local, and
+# this one writes to the remote. Run it once `make release` has finished and
+# the tag exists.
+VERSION := $(shell awk -F'"' '/MARKETING_VERSION:/ {print $$2}' project.yml)
+TAG     ?= v$(VERSION)
+
+publish: $(DMG)
+	@test -n "$(VERSION)" || (echo "No MARKETING_VERSION in project.yml" && exit 1)
+	@# --clobber so re-running after a rebuild replaces the asset instead of
+	@# failing on the name already being taken.
+	gh release upload $(TAG) $(DMG) --clobber
+	@echo "Attached $(DMG) to $(TAG)."
 
 # What Gatekeeper on a customer's Mac will check. `spctl` accepting the app is
 # the actual proof that the download will open without a right-click.
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ two never disagree.
 
 | Provider | Source | How |
 |---|---|---|
-| **Claude Code** | official | The OAuth token in the login keychain, against the same endpoint Claude Code's own `/usage` uses. |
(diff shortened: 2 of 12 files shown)


mulhamna:
@vinzdg — two offers, both easy to decline.

**On the size of this PR.** It carries four separable things: the signing fix, the release asset, Claude's source order, and Antigravity's. I bundled them because they were found in one sitting, but that means disagreeing with one part costs you all four. Say the word and I'll split it — the work is already sitting on three separate branches locally, so it is a push, not a rewrite. The signing fix in particular is one character and unblocks anyone trying `make build`; there is no reason for it to wait behind the provider changes if those need discussion.

**On releases.** There is no `.github/` in the repo today, so every check is manual. I am happy to contribute a workflow, and it is worth being clear about what CI can and cannot do here before you decide:

1. **A test workflow** — `make test` on pull requests. No secrets, nothing to configure. This is the part with real value and no cost; right now nothing verifies a contribution before you look at it.
2. **`release-please`, driven by `workflow_dispatch`** so a release only ever happens when you press the button — it would take the version bump, the `CHANGELOG`, the tag and the GitHub release off your hands. The honest catch: it reads Conventional Commits, and this repo's history is mixed. Your own commits read `Codenotch 1.5.0` and `Add OpenCode as a provider`, while `fix: …` mostly arrives through PRs. Adopting it means committing to that style yourself, which is a real change to how you work and not obviously worth it for a project this size. Your call entirely.
3. **Notarizing in CI** — possible, but it means putting your Developer ID certificate and an app-specific password into repository secrets. I would not suggest it. Keeping `make release` on your machine and running `make publish` after it gets the dmg onto the release page with nothing sensitive leaving your Mac, which is what this PR does.

So: 1 is nearly free and I would do it regardless; 2 only if the commit convention appeals to you; 3 probably never. Tell me which, if any, and I will open it separately from this one.

Two things in this PR I could not verify myself, both flagged in the description and worth repeating here: `make publish` has never actually been run, because it needs write access to this repository, and the Antigravity path is covered only by unit tests, because Antigravity is not installed on the machine I wrote this on. If either matters to you before merging, I would rather you knew now than found out after.

vinzdg:

Claude now asks its own CLI's `/usage` first and only falls back to the keychain, which removes the repeated keychain prompts.

Genuinely appreciated.

PR #24, opened 2026-09-07
Make `make` actually sign a Debug build without a Developer ID

florianalber:
## The bug

The ad-hoc fallback added for 1.5.0 ("Contributors can now `make build`/`make test` with no Apple Developer account") never engages. It is guarded by

```make
```

and `grep -c` prints `0` rather than nothing when it matches nothing — so the condition is false on **every** machine, including the ones it exists for. `DEV_SIGN` stays empty, the build falls through to `project.yml`'s manual identity and team `6WFPL8B9FB`, and anyone without that certificate gets

```
error: Signing for "Codenotch" requires selecting either a development team or
a provisioning profile.
```

from `make build`, `make test` and `make run` alike. On the maintainer's own machine nothing looks wrong, because there the fallback is supposed to stay empty.

## The change

Dropping `-c` is the whole fix — `grep` alone prints nothing when it matches nothing, which is what `ifeq (,…)` is testing for.

The ladder above it is one step wider, and separable if you'd rather not have it: where the machine has a personal **Apple Development** certificate, that is now preferred over ad-hoc, for exactly the reason `project.yml` pins a stable identity on yours. The keychain ACL that guards Claude Code's and Antigravity's credentials keys on the signing identity, and an ad-hoc identity is new on every build — so "Always Allow" is forgotten and the prompt comes back after every `make run`, which is painful precisely when working on the credential-reading paths. Manual signing will not proceed without a team, so the team is read out of the certificate's own subject; if either half comes back empty, ad-hoc still carries a build with no Apple account at all.

If you prefer the minimal version, say so and I'll cut this back to the one-flag change.

## Verified

On a Mac with an Apple Development certificate and **no** Developer ID:

- `make build` → signs manually against the team parsed from the certificate, **BUILD SUCCEEDED**
- `make build DEV_TEAM=` (the certificate withheld, so the ad-hoc branch runs) → **BUILD SUCCEEDED**, and the product comes out `Signature=adhoc`, `TeamIdentifier=not set`
- `make test` → 552 tests, 0 failures (macOS 27, Xcode 26.6)

Before the change, all three of those failed at the signing step on the same machine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,38 @@ PROJECT := Codenotch.xcodeproj
 SCHEME  := Codenotch
 DEST    := platform=macOS,arch=arm64
 
-# Debug ad-hoc signs itself when the maintainer's Developer ID certificate
-# isn't in the keychain, which is every machine but the maintainer's — so a
-# contributor can `make build`/`make test`/`make run` with no Apple account at
-# all, per CONTRIBUTING.md. On the maintainer's own machine this is empty and
-# changes nothing: project.yml's stable identity is what keeps a keychain
-# "Always Allow" grant alive across rebuilds, and forcing ad-hoc there would
-# throw that away and bring the prompt back on every `make run`.
-ifeq (,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
+# Debug signs itself when the maintainer's Developer ID certificate isn't in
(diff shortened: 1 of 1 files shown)


vinzdg:

Debug builds now sign themselves when the maintainer's Developer ID isn't in the keychain, so contributors can build with no Apple account.

Genuinely appreciated.

PR #25, opened 2026-09-07
Read the default profile's suffixed keychain service, not just the bare name

at10249:
Fixes the Claude default-profile ring being stuck on **"Waiting for the first reading…"** on machines where `CLAUDE_CONFIG_DIR` is exported.

### Symptom
On my Mac the Claude usage ring never got a reading — "Waiting for the first reading…" indefinitely — while **session/idle detection worked fine**, and Codex (and others) read normally. That contrast is the tell: session detection reads `~/.claude/sessions` files, but the usage number needs the OAuth token from the keychain.

### Root cause
Claude Code files its OAuth token under the **SHA-256-suffixed** service whenever `CLAUDE_CONFIG_DIR` is set in the shell it runs from — and a shell that exports the variable exports it **even when it points at the default `~/.claude`**. So the default profile's live, hourly-rotating token was under `Claude Code-credentials-<hash of ~/.claude>`, while the bare `Claude Code-credentials` held only a stale duplicate from months earlier (before the variable was exported).

`ClaudeProfile` read only the bare name for the default profile (`slug == nil`), so it picked the expired duplicate → `GET /api/oauth/usage` returned `401` → the ring waited forever, with a valid token one service name away.

Concrete evidence on my machine: `sha256("/Users/<me>/.claude")[:8]` matched an existing keychain item `Claude Code-credentials-<hash>` whose token was current, while the bare item had expired 42 days earlier.

### Fix
Give the default profile **both** candidate services — suffixed first, bare kept as a legacy fallback — and let `KeychainItem` pick the newest item **across** them. This reuses the existing "newest wins" philosophy you already apply to rotation duplicates *within* one service, just widened across service names. Named profiles are unchanged (single suffixed service, no bare fallback that could shadow another account).

- `ClaudeProfile`: `keychainService` → `keychainServices` (`[suffixed, bare]` for the default; `[suffixed]` for named). `keychainService` retained as the first candidate for existing callers/tests.
- `KeychainItem`: add `newest(services:)` / `modifiedAt(services:)`.
- `ClaudeCredentials` / `ClaudeKeychain`: read across the candidate list; the shared default reader uses the default profile's full list.
- Tests: assert the default's two-service candidate list and the named profile's single service.

### Testing
Reasoned against the codebase and syntax-checked with `swiftc -parse`; I'm on macOS 15 so I couldn't run the full `make test` locally (Xcode 26 required). Happy to adjust naming/placement to fit your conventions.

### DIFF
--- a/Sources/Providers/ClaudeCredentials.swift
+++ b/Sources/Providers/ClaudeCredentials.swift
@@ -42,9 +42,9 @@ struct ClaudeCredentials {
     /// authorization to read — only this second, targeted fetch of the
     /// winner's actual data does, which is why it costs the same single prompt
     /// as before, per profile.
-    static func read(service: String) throws -> ClaudeCredentials {
-        guard let winner = KeychainItem.newest(service: service) else {
-            Log.usage.error("keychain read failed: no item under \(service, privacy: .public)")
+    static func read(services: [String]) throws -> ClaudeCredentials {
+        guard let winner = KeychainItem.newest(services: services) else {
+            Log.usage.error("keychain read failed: no item under \(services.joined(separator: ", "), privacy: .public)")
             throw UsageProviderError.needsAuth
         }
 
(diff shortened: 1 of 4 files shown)


vinzdg:

The default profile now offers the suffixed keychain service as well as the bare name, so a live token filed under the hashed service is found.

Genuinely appreciated.

PR #26, opened 2026-09-07
Show provider usage in the menu bar menu

MintarasGrinius:
Settings has a Hide option for the notch overall, so putting the same usage on the menu bar click feels natural: Hide plus App icon set to Menu bar becomes a working setup instead of a one-way door to numbers you can no longer see.

Outcome (screenshot to attach below):

The menu bar menu now lists every provider with the same facts as the notch tooltips — header row with headline figure and reading age (clicking it refreshes just that provider), one line per limit window (label, percentage, reset), the blocked line first when present, or the status message when nothing is metered. Menu is rebuilt on every open so countdowns and ages are fresh, followed by Refresh all, Settings…, Quit. The Hide explanation now points at the menu bar route.

[Screenshot: menu bar usage menu — drag image 1 here]

Also in this PR, two small fixes found along the way:

- Makefile: the ad-hoc signing fallback never triggered — `grep -c` prints `0`, never empty, so `ifeq (, …)` was never true and Debug builds demanded the maintainer's Developer ID cert. Now `ifeq (0, …)`.
- Codex: one window with a null `used_percent` threw the whole fetch away, hiding a good weekly window behind a bad hourly one (ring showed — with only Weekly in the tooltip). Malformed windows are now skipped; both-bad is still nothingMetered. `limit_window_seconds` is optional with a fallback label.

Tests: `make test` passes — 561 tests, 0 failures, including 6 new StatusMenuTests and 3 new Codex partial-window tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ DEST    := platform=macOS,arch=arm64
 # changes nothing: project.yml's stable identity is what keeps a keychain
 # "Always Allow" grant alive across rebuilds, and forcing ad-hoc there would
 # throw that away and bring the prompt back on every `make run`.
-ifeq (,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
+ifeq (0,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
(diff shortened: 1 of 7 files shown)


MintarasGrinius:
Screenshot 2026-09-07 at 12 00 23

vinzdg:

The menu bar item now lists each provider's usage, not just a way into Settings.

Genuinely appreciated.

PR #28, opened 2026-09-07
Allow pinning the notch to a display

### LINKED ISSUE #16: Notch jumps between displays on a multi-monitor Mac — a way to pin it to one screen
## Setup

- Codenotch 1.4.1
- macOS 26.6.2
- Three displays: a 3440×1440 ultrawide (the main display in System Settings), the built-in Retina display, and a 1920×1080 monitor

## What happens

The notch moves to whichever display holds the window I just focused. Across three screens that means it is somewhere different every few minutes, so glancing at it — the one thing it is for — stops being a reflex and becomes a search.

## Where it comes from

`NotchGeometry.preferredScreen` picks `NSScreen.main`:

```swift
/// The notch follows the screen with the menu bar.
static func preferredScreen(from screens: [NSScreen]) -> NSScreen? {
    NSScreen.main ?? screens.first
}
```

The comment says "the screen with the menu bar", but `NSScreen.main` is documented as the screen containing the window that is currently receiving keyboard events — the focused window's screen, not the menu bar's. On a one-display Mac the two are the same object, so the difference only surfaces with more than one screen.

## What I'd like

A display choice in Settings, next to the existing edge picker: "Follow the active window" for today's behaviour, plus the option to pin the notch to one named screen.

If the intended behaviour really is the menu-bar screen, then resolving that screen explicitly — the one whose `frame.origin` is `.zero` — would match the comment and already be stable for most multi-display setups.

Happy to test a build.

RrezonCurraj:
The notch currently follows `NSScreen.main`, so it moves whenever keyboard focus moves to a window on another display. This adds a Display picker beside the existing edge controls, allowing the notch to keep following the active window or stay pinned to a named physical display.

The selected display is stored by its persistent macOS display UUID. If that display is disconnected, Codenotch temporarily follows the active window and automatically returns when the selected display reconnects.

Closes #16.

## Validation

- `make test` — 557 tests, 0 failures
- Verified the real Settings UI lists “Follow active window” and connected displays by name
- Verified changing the selection updates and persists the preference
- Covered selected, active, disconnected, and reconnected display resolution with simulated screens

I only have one physical monitor available, so the final movement between physical displays still needs hardware verification. @makifej, would you be willing to test this draft on your three-display setup?

### DIFF
--- a/Sources/App/AppDelegate.swift
+++ b/Sources/App/AppDelegate.swift
@@ -81,6 +81,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             // this, every launch on any other edge opens with a flash of the
             // right-hand one and then crossfades away from it.
             controller.model.edge = preferences.notchEdge
+            controller.displayPreference = preferences.displayPreference
 
             let updater = Updater()
             self.updater = updater
@@ -144,6 +145,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             preferences.$notchEdge
                 .receive(on: RunLoop.main)
                 .sink { [weak controller] in controller?.apply(edge: $0) }
+                .store(in: &cancellables)
+
+            preferences.$displayPreference
+                .receive(on: RunLoop.main)
+                .sink { [weak controller] in controller?.apply(displayPreference: $0) }
                 .store(in: &cancellables)
 
             preferences.$disconnectedProviders
--- a/Sources/Notch/NotchGeometry.swift
+++ b/Sources/Notch/NotchGeometry.swift
@@ -14,16 +14,29 @@ protocol ScreenDescribing {
     var frameValue: CGRect { get }
     var visibleFrameValue: CGRect { get }
     var hardwareNotch: HardwareNotch? { get }
+    var displayIdentifier: String? { get }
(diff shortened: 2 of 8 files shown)


makifej:
Happy to test this on the three-display setup.

One blocker: I don't have Xcode installed here, only the Command Line Tools, so `make build` won't run on my machine — and installing Xcode just for this is more than I can take on right now.

Could you or @vinzdg put a built `.app` or `.dmg` from this branch somewhere I can download? I'll run it as soon as there's a build.

Once I have one, here's what I'd check:

- The Display picker lists all three monitors by name
- Pinned to one display, the notch stays put while I focus windows on the other two
- The choice survives an app restart
- Unplugging the pinned display — does it fall back to following the active window, and does the picker keep the selection rather than silently resetting to "Follow active window"?
- Replugging it — does the notch return on its own, without opening Settings?
- Rearranging displays in System Settings (changing which one is primary) doesn't break the pin

Let me know if there's anything else you'd like covered.

vinzdg:

Appearance has a Display picker that pins the notch to a named screen, and says so when that screen is disconnected.

Genuinely appreciated.

PR #30, opened 2026-09-07
Say when a session ends, and jump to it

philharmonie:
The monitors already know the moment an agent stops working — the notch just never said so, which leaves it as something to watch rather than something that tells you. This closes that gap: a busy session going idle, or going to waiting, opens the notch for a few seconds and plays a sound, and a click while it is open raises the application that session is running in.

### What it does not announce

- **Waiting → idle.** A question that was answered is not a piece of work ending; announcing it would fire on every prompt.
- **A session whose file disappears mid-turn** — what quitting Claude Code looks like. There is no window left to jump to.
- **The first reading.** Every session already running at launch arrives with no history, and treating that as a transition would ring once per open window on every start, including a restart in the night after an update.

`SessionCompletionWatcher` holds all of that, free of AppKit so the edge cases are driven directly from tests. `AgentSession` gains an optional pid — Claude Code and Grok publish one, the other three monitors do not, and a nil there costs only the ability to jump.

### The jump raises the app, not the tab

A session publishes a pid and nothing else — no window, no tty — so `SessionFocus` walks the process tree from the agent to whatever launched it. Selecting the *tab* inside that app needs the terminal's own scripting interface and there is no general one: Terminal.app and iTerm2 can match a tab by tty, Warp and Ghostty publish no scripting dictionary at all. Raising the app for everybody, with the tooltip naming the session, beat working in two terminals and silently doing nothing in a third.

### Two things that are not obvious from the diff

**A peek is not the pointer arriving, so the pointer leaving must not end it.** `cursorMoved` runs every 0.3s and folds the notch 0.45s after finding the pointer elsewhere, which during a peek it almost always is. Without a deadline the fold logic respects, the notch opened and shut inside a second and the announcement could not be seen. `PeekTests.testAPeekOutlastsTheHoverFold` covers it — I checked it fails with the guard removed.

**The chime plays the file through `AVAudioPlayer` rather than handing a name to `NSSound`.** `NSSound(named:)` resolves a system alert, and system alerts route through the interface sound-effects channel — the one System Settings → Sound switches off with \"Play user interface sound effects\", which is also what makes the Mac click and swoosh all day, so plenty of people have it off. On my machine it is off (`com.apple.sound.uiaudio.enabled = 0`), and I would rather an announcement not depend on a setting that is about something else entirely. Reading the file into an `AVAudioPlayer` puts it on the ordinary output path, where the only thing that silences it is the volume control. `NSSound` stays as the fallback for a file AVFoundation will not open.

To be straight about it: I first reached for this after a silence that turned out to be my own bug — `play()` written inside the `Logger` interpolation reporting its result, which on an ordinary run is never evaluated. That is fixed and tested here (`testPlayingActuallyStarts`), and I have **not** since verified that `NSSound` alone would be inaudible with the sound-effects channel off. The reason above is the mechanism and the setting, not a measurement. Happy to go back to `NSSound` if you would rather keep the dependency small.

### Settings

Its own section rather than a line in General — this is the only part of the app that speaks first, and a switch that stops the Mac making a noise has to be findable by someone looking for exactly that.

The peek and the sound switch off separately because they fail differently: the peek is no use behind a full-screen window, the sound is no use in a meeting. Each of the two events picks its own sound with a preview button beside it, since choosing an alert you cannot hear until the next time it fires is guesswork. The peek's length is a choice of three — below a second or two it is gone before a glance lands, much beyond ten it stops reading as an announcement.

### Checks

- `make test` passes: 578 tests, 0 failures.
- 25 new tests across `SessionCompletionWatcher`, `SessionFocus`, `PeekDuration` and the chime, in `Tests/SessionCompletionTests.swift`. Two of them are regression tests for bugs found while building this, and I checked both fail with their fix removed: the peek being folded away by the cursor poll, and `play()` being swallowed by a `Logger` autoclosure.
- No layout constants touched, so nothing to re-check against the design frame.

Verified against real sessions as well as fixtures — the unified log shows the transition, the chime and the peek's full length:

```
session codenotch-selftest finished
chime Glass: true
peek for 5.000000s, pid 91202
peek folded          ← 5.02s later
```

Note: `make build` currently fails on a machine without a Developer ID certificate, for a reason unrelated to this change — the ad-hoc fallback in the Makefile never fires. That is #24's to fix; this PR was built and tested with the override applied by hand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,38 @@ Any `~/.claude-<slug>` directory Claude Code has run against is found at launch;
 the default `~/.claude` always comes first, the rest in alphabetical order, so the
 rings never swap places.
 
+## When a session ends
+
+The notch opens itself for five seconds when an agent stops working, or stops
+to ask you something, and sounds the system alert. Clicking it while it is open
+brings that session's application to the front.
+
+The app, not the tab. A session publishes its pid and nothing else — no window,
+no tab, no tty — so the app is found by walking up the process tree from the
+agent to whatever launched it. Choosing the *tab* inside that app needs the
+terminal's own scripting interface, and there is no general one: Terminal.app
+and iTerm2 can match a tab by tty, Warp and Ghostty publish no scripting
+dictionary at all. So the app is raised for everybody and the tooltip names the
+session, which leaves the last hop one keystroke rather than working for two
+terminals and silently doing nothing in a third.
+
+Both halves switch off separately in Settings, because they fail differently:
+the peek is no use behind a full-screen window, and the sound is no use in a
+meeting. Each of the two events — finished, and waiting on you — picks its own
+sound there, with a preview button beside it.
+
+The sound is played as a file on the ordinary output rather than handed to
+`NSSound` as a system alert. A system alert goes through the interface
+sound-effects channel, which System Settings → Sound can switch off — and on a
+Mac where it is off, `NSSound.play()` reports success and nothing is heard.
+
+Only *leaving* busy counts. A question being answered is not a piece of work
+ending, and a session whose file disappears mid-turn — which is what quitting
+Claude Code looks like — is not announced at all, since there is no window left
+to jump to. Nothing is announced from the first reading either: every session
+already running at launch arrives with no history, and treating that as a
+transition would ring once per open window on every start.
+
 ## Placement
 
 The notch lives on any of the four screen edges. Right and left keep a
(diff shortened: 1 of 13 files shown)


vinzdg:

The notch now opens itself for a few seconds and sounds a chime when an agent stops working or starts waiting on you; clicking it raises that session's app.

Genuinely appreciated.

PR #31, opened 2026-09-07
Fix the ad-hoc signing fallback, which never fired

philharmonie:
CONTRIBUTING says `make build` needs no Apple Developer account, and the comment above the test in the Makefile says the same: with no Developer ID certificate in the keychain, `DEV_SIGN` should force ad-hoc signing so a contributor can build. It never did.

The test asks whether `grep -c` produced empty output:

```make
```

`grep -c` prints a count, and the count on no match is the string `"0"` — never empty. So the `ifeq` was false on exactly the machines it exists for, `DEV_SIGN` stayed empty, and the build used `project.yml`'s identity:

```
error: No signing certificate "Developer ID Application" found
```

Comparing against the count instead fixes it. On the maintainer's machine, where the count is `1`, nothing changes.

Verified both ways on a Mac with no Developer ID certificate: `make build` fails on `main` and succeeds on this branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,11 @@ DEST    := platform=macOS,arch=arm64
 # changes nothing: project.yml's stable identity is what keeps a keychain
 # "Always Allow" grant alive across rebuilds, and forcing ad-hoc there would
 # throw that away and bring the prompt back on every `make run`.
-ifeq (,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
+# `grep -c` prints "0" and exits 1 when it matches nothing, so the emptiness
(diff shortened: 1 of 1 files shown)


philharmonie:
Closing as a duplicate — #24 found this first and fixes it better: it drops `-c` entirely rather than comparing against "0", and additionally prefers an existing "Apple Development" certificate over ad-hoc, with the team parsed out of the certificate. That keeps a keychain "Always Allow" grant alive across rebuilds, which matters for anyone working on the credential-reading paths. #35 is the same one-line change as this one.

Leaving it to #24. Sorry for the noise — I hit the build failure before reading the open PRs.

PR #32, opened 2026-09-07
Fix: allow dotted AM/PM markers in ResetCopyTests on non-English locales

divingsbysangam:
## Problem

`ResetCopyTests.testAbsoluteTimeUsesAColon` asserted `!text.contains(".")`. The test was written to verify that the time separator is a colon (`h:mm`) rather than a period (`h.mm`), which occurs in some European templates.

However, on Spanish (and several other non-English) locales, AM/PM markers format with periods as `"a. m."` / `"p. m."`. As a result, running `make test` on a Spanish-locale Mac failed `testAbsoluteTimeUsesAColon` even though the time separator was correctly formatted as a colon. (This pre-existing test failure was also noted in #20).

Additionally, `ResetCopy.formatter(for:)` unconditionally hardcoded `formatter.locale = .current`, ignoring any custom `calendar.locale` that callers or test suites injected.

## Changes

1. **`Sources/Model/ResetCopy.swift`**:
   - `ResetCopy.formatter(for:)` now respects `calendar.locale ?? .current`.
2. **`Tests/ResetCopyTests.swift`**:
   - Updated `testAbsoluteTimeUsesAColon` to check that no period occurs between digits (`#"\d+\.\d+"#`), keeping the colon time-separator check while permitting dotted AM/PM markers.
   - Added `testSpanishLocaleKeepsColonTimeSeparator` to explicitly pin and verify this behavior under `es_ES`.

## Verification

- `make test` passes with 0 failures (553 passed tests).

### DIFF
--- a/Sources/Model/ResetCopy.swift
+++ b/Sources/Model/ResetCopy.swift
@@ -47,7 +47,7 @@ enum ResetCopy {
         let formatter = DateFormatter()
         formatter.calendar = calendar
         formatter.timeZone = calendar.timeZone
-        formatter.locale = .current
+        formatter.locale = calendar.locale ?? .current
         return formatter
     }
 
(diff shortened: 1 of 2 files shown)


vinzdg:

`ResetCopy` keeps a colon time separator regardless of locale, and there's a Spanish-locale test pinning it.

Genuinely appreciated.

PR #33, opened 2026-09-07
Show a notch on every display or just the main one

MintarasGrinius:
Appearance gets a Displays picker: Main display / All displays. All displays puts a notch on every connected screen, each with independent hover — reaching for one opens only that one.

Why one notch per display: each panel is welded to its own screen edge with its own open state and screen size (tooltip budgets come out of it). A shared model would unfold every notch when the pointer reaches any of them, including the display you are not looking at.

- New `NotchScreenScope` setting (`notchScope`, default main display so current behavior is unchanged) plus a Displays segmented picker next to Edge.
- New `NotchFleet`: owns one `NotchWindowController` per display keyed by `NSScreenNumber`, reconciles on screen-parameter changes, and fans out snapshots, sessions, refreshing, edge, visibility and callbacks. A controller created late (display plugged in at noon) starts with current readings. The single-notch case keeps its controller and moves it, the way a lone panel always did.
- `NotchWindowController` gains `assignedScreen` (nil follows the menu-bar screen, i.e. old behavior) and `retire()` for clean teardown when a display disconnects.

Tests: 564 tests, 0 failures — 12 new (scope defaults/persistence/unknown-value fallback, pure reconcile maths, fleet counts against the real screen list, late-controller readings). No layout-math changes.

Notes: merge after #26 — plain `make test` on a contributor machine needs its Makefile ad-hoc fix (verified here with manual ad-hoc flags). And the all-screens path is test-exercised only; I have one display, so worth an eyeball on a multi-screen setup.

### DIFF
--- a/Sources/App/AppDelegate.swift
+++ b/Sources/App/AppDelegate.swift
@@ -3,7 +3,7 @@ import Combine
 import SwiftUI
 
 final class AppDelegate: NSObject, NSApplicationDelegate {
-    private var notchController: NotchWindowController?
+    private var notchFleet: NotchFleet?
     private var store: UsageStore?
     private var monitors: [String: any AgentActivityMonitor] = [:]
     private var preferences: Preferences?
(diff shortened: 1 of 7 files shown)


vinzdg:

Appearance has a Displays picker — the main display, or a notch on every one — with one controller per screen behind it.

Genuinely appreciated.

PR #34, opened 2026-09-07
Fix Cursor enterprise usage metering and Auth0 sign-in detection

Minitour:
## Summary
- Parse enterprise/team `individualUsage.overall` (`used`/`limit`) so the Cursor ring shows real usage instead of \"enterprise plan has nothing to meter\".
- Fall back to the access-token JWT `sub` when `cursorAuth/stripeMembershipAuthId` is missing (common on Auth0/enterprise), which previously showed \"sign in to Cursor in the IDE\" while already signed in.
- Add pinned tests from a live enterprise `GET /api/usage-summary` response.

## Test plan
- [ ] `make test`
- [ ] On an enterprise Cursor account: ring shows ~used/limit (e.g. 6907/45000 ≈ 15%) and resets at billing cycle end
- [ ] On a free/pro account: existing `plan.totalPercentUsed` behavior unchanged
- [ ] With `stripeMembershipAuthId` deleted from `state.vscdb` but a valid `accessToken` present: Cursor still authenticates

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Sources/Providers/CursorCredentials.swift
+++ b/Sources/Providers/CursorCredentials.swift
@@ -55,15 +55,40 @@ struct CursorCredentials {
         defer { sqlite3_close(db) }
 
         guard let token = value(forKey: "cursorAuth/accessToken", in: db),
-              let account = value(forKey: "cursorAuth/stripeMembershipAuthId", in: db),
-              !token.isEmpty, !account.isEmpty
+              !token.isEmpty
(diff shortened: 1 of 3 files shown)


vinzdg:

Auth0/WorkOS sign-in detection landed from this, alongside the enterprise/team usage shape.

Genuinely appreciated.

PR #36, opened 2026-09-07
feat: Grok Bot, GitHub Copilot and OpenCode Go providers; threshold alerts, notch control, CI

virajshoor:
## What this adds

Three new usage providers, threshold alerts, notch control, and CI — on top of the foundation that already exists.

### New providers (all `.official`, all borrowed credentials, response shapes pinned by tests)

- **Grok Bot** — reads the installed desktop app's active account and selected team via its own `DashboardService` usage status. Pooled enterprise allowances are reported honestly rather than turned into an invented percentage.
- **GitHub Copilot** — GitHub's Copilot quota endpoint, authenticated with the GitHub CLI session already on the Mac (`gh auth login`). Unlimited quotas are dropped rather than shown as zero.
- **OpenCode Go** — the rolling 5-hour, weekly and monthly dollar-value limits from OpenCode's own quota endpoint, with the API key the OpenCode desktop app stores in its shared auth file. The endpoint is undocumented, so its shape is pinned by tests as the first place a change would show.

### Threshold alerts

A provider's headline limit crossing **80%** — and reaching **100%** — becomes a system notification: once per crossing, never repeated while it stays crossed, and again only after the window rolls over. Muted per provider from its settings row; macOS permission is requested on the first real crossing, not at launch.

### Notch control

- **Hide for an hour** in the context menu — a timed reprieve that expires by itself, reversible from Settings' Appearance section. Deliberately *not* the standing Hidden choice.
- **Provider reordering** from the settings rows, persisted; ids missing from a saved order trail in natural position, so a provider added later appears rather than vanishing behind a stale arrangement.

### Fractional percentages

Below one percent, whole percents collapse a real reading into "0%" — the one number that looks most like *nothing used*. Sub-1% readings now show a tenth of a percent ("0.3%"), both halves move together so they still add up, and a tenth of nothing says "<0.1%" rather than pretending to be zero. Above one percent, nothing changes.

### Refactors and CI

- The tooltip card's height was assembled from a snapshot at three call sites (the card, the hover region, the placement centre); the arguments now live in one `NotchLayout.cardHeight(for:)` so the three cannot drift apart.
- Animation timings that had drifted out of `NotchMotion` move back in; `ClaudeSessionMonitor.start()` is idempotent.
- Tests for `ProcessLiveness` (pid reuse tolerance) and the defaults-domain migration — 588 total, all passing locally.
- GitHub Actions workflow running the suite on every push and PR (`make test-ci`, unsigned — the tests read fixtures, not keychains).

## Verification

- `make test`: **588 tests, 0 failures** on the commit head.
- Secrets audit of the full branch diff: no credentials, tokens, private keys or notary material — the Sparkle EdDSA *private* key and notary profile are keychain-only and stay there; `SUPublicEDKey` is public by design.

## Credits

Codenotch is **Vinz**'s ([@hivinz_](https://x.com/hivinz_)) — the design, the notch, and every idea that makes it work are his. This branch adds providers, alerts and polish on top of that foundation. The MIT license and original copyright are preserved.


vinzdg:

The GitHub Copilot ring, the 80%/100% threshold alerts and the CI workflow all landed from this. Grok and OpenCode had already gone in separately by then.

Genuinely appreciated.

PR #39, opened 2026-09-07
Add a Gemini API key provider read from the tools' own logs

indh-x:
## Summary

A bare `GEMINI_API_KEY` is billed per token and has no usage endpoint, so this adds a `gemini-api` ring ("Gemini API", Gemini sparkle glyph) that sums the per-call token counts three local tools already record for that key:

- **Gemini CLI** — `~/.gemini/tmp/<project>/chats/session-*.jsonl` (`tokens.total` per `gemini` record; records are appended twice and deduped by `id`, `$set`/`$rewindTo` ignored)
- **OpenCode** — `~/.local/share/opencode/opencode.db`, `message.data` with `providerID == "google"` (`tokens.total`, or the component sum for aborted rows)
- **Hermes** — `~/.hermes/state.db`, `session_model_usage` with `billing_provider == "gemini"` (`input + cache_read + cache_write + output`; reasoning already sits inside output)

The key itself is never read. Windows: current local month (headline) and today, plus one row per tool that has state on disk. Fidelity is `.derived`; setting a monthly token budget in Settings turns the count into a ring fraction and flips it to `.manual`. Gemini CLI session-file mtime drives busy/idle. Counts ≥ 10k render compact (`651k`, `1.1M`).

The sparkle gets a new glyph key `gemini-spark` because the raw value `gemini` already names the Antigravity arch in archived snapshots. Design notes and the rejected Cloud Monitoring path are in `TASKS.md`.

Deliberately not done: version bump / release note (maintainer's release step), USD estimates, any network call.

Stacked on #35 (`fix-adhoc-signing-check`), whose commit this branch includes so contributors can build; rebases cleanly once #35 lands.

## Test plan

- [x] `make test` — 605 tests, 0 failures (52 new: readers against inline JSONL and WAL SQLite fixtures mirroring the real schemas, snapshot builder, budget round-trip, monitor, glyph geometry)
- [x] Readers' SQL executed read-only against real `opencode.db` and `state.db`
- [x] `make run` — Gemini API ring (sparkle) next to Antigravity (arch) and the budget field in Settings, verified visually by the author

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ DEST    := platform=macOS,arch=arm64
 # changes nothing: project.yml's stable identity is what keeps a keychain
 # "Always Allow" grant alive across rebuilds, and forcing ad-hoc there would
 # throw that away and bring the prompt back on every `make run`.
-ifeq (,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
+ifeq (0,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
 DEV_SIGN := CODE_SIGN_IDENTITY="-" DEVELOPMENT_TEAM="" CODE_SIGN_STYLE=Automatic
 endif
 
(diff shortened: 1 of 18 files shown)


vinzdg:

A Gemini API ring reads the key the tools already hold, with a monthly token budget set in Settings since a bare API key publishes no limit.

Genuinely appreciated.

PR #40, opened 2026-09-07
Let the pill be nudged along its edge with an ⌥-drag

Slayther13:
## Summary

The notch is always centred on its edge, with no way to move it — which collides with any other menu-bar-adjacent app anchored to the same spot (this is what prompted the change).

- ⌥-drag on the pill now nudges it along the edge live.
- The offset is clamped so it can't be dragged off screen, and is remembered per edge (right/left/top/bottom) across restarts.
- A plain click, without ⌥, behaves exactly as before — open/pin/refresh a provider/open Settings.

## Why the clamp needed its own logic

`panelSize` budgets enough room on each end for the tallest possible hover card, so the panel itself is close to the full screen height/width on most screens. Clamping the drag by that padded size left only a handful of points of travel. The clamp now uses `NotchViewModel.slack` — the padding `panelSize` carries beyond the visible pill — so only the pill itself has to stay on screen; the invisible hover-card padding is free to run past the bezel.

## Test plan

- [x] `make test` — all 557 tests pass, including new coverage in `NotchGeometryTests.swift` for the offset's sign convention (down/right is positive) and for the slack-based clamp.
- [x] Manually verified on a MacBook: ⌥-drag on all four edges, offset persists across a relaunch and is independent per edge, plain clicks still open/pin/refresh as before.

### DIFF
--- a/Sources/App/AppDelegate.swift
+++ b/Sources/App/AppDelegate.swift
@@ -81,6 +81,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             // this, every launch on any other edge opens with a flash of the
             // right-hand one and then crossfades away from it.
             controller.model.edge = preferences.notchEdge
+            controller.model.alongOffset = preferences.offset(for: preferences.notchEdge)
 
             let updater = Updater()
             self.updater = updater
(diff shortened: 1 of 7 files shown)


vinzdg:

The pill takes an Option-drag along its edge now, and the nudge is remembered per edge.

Genuinely appreciated.

PR #41, opened 2026-09-07
Add customizable accent color setting

aarushbhadragiri:


vinzdg:

Appearance has an accent colour picker. The palette grew to ten fixed colours plus the device accent.

Genuinely appreciated.

PR #43, opened 2026-09-07
feat: add a reset time format preference

TrapstarKS:
Reset times currently switch from a countdown under an hour to a weekday or date for longer waits. Users who prefer knowing how much time remains have to work that out themselves.

This change adds a Reset time preference under Settings → Appearance:

- Choose Time remaining to show countdowns such as "Resets in 3 Days 3h" or "Resets in 3h 20m".
- Keep Reset date as the default, preserving the existing display, including minutes when less than an hour remains.
- Save the preference across app restarts and apply changes immediately to usage tooltips.
- Reuse the existing 30-second refresh to keep countdowns current.

Validation:

- Added coverage for countdown formatting, rounding across hour and day boundaries, expired resets, preference persistence, and unknown stored values.
- Full suite: 554 tests passed on Apple silicon.
- Rebuilt and launched the app locally after the final label change to Reset date.
- `git diff --check` passed.

The full suite was run from a temporary source copy after Xcode stalled while reading the working directory, using local signing overrides because the maintainer's certificate was unavailable:

```sh
make test 'DEV_SIGN=CODE_SIGN_IDENTITY="-" DEVELOPMENT_TEAM="" CODE_SIGN_STYLE=Automatic'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ At rest it is a small pill on the screen edge that unfolds when the pointer
 reaches it — configurable in Settings to always show, or to hide entirely.
 Settings live in an orb below the notch: an arc at rest, a gear on hover.
 
+In Settings → Appearance → Reset time, choose **Time remaining** for countdowns
+like "Resets in 3 Days 3h". **Automatic** keeps the reset date and time, with
+minutes shown when less than an hour remains.
+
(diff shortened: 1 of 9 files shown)


vinzdg:

Appearance's Reset time picker switches between a countdown and a reset date.

Genuinely appreciated.

PR #46, opened 2026-09-08
fix: ad-hoc signing fallback never triggers in Makefile

### LINKED ISSUE #45: Makefile: ad-hoc signing fallback never triggers (grep -c never outputs empty)
## What happens

On a non-maintainer machine with zero \"Developer ID Application\" identities in the keychain, `make build` / `make run` / `make test` fail with:

```
error: No signing certificate "Developer ID Application" found: No "Developer ID Application" signing certificate matching team ID "6WFPL8B9FB" with a private key was found.
```

even though the Makefile and CONTRIBUTING.md both say this should just work with no Apple Developer account.

## Root cause

```make
DEV_SIGN := CODE_SIGN_IDENTITY="-" DEVELOPMENT_TEAM="" CODE_SIGN_STYLE=Automatic
endif
```

`grep -c` always prints a count, even when there are zero matches — it prints the string `0`, not an empty string. So `$(shell ... | grep -c ...)` is never empty, and `ifeq (,...)` is never true. `DEV_SIGN` is therefore never set to the ad-hoc fallback, regardless of whether a Developer ID certificate is actually present.

Confirmed on a clean machine:

```
$ security find-identity -v -p codesigning
    0 valid identities found
```

yet `grep -c "Developer ID Application"` on that output still returns `0`, not empty, so the `ifeq` guard fails to fire.

## Suggested fix

Compare against `0` instead of empty string, e.g.:

```make
ifeq (0,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
DEV_SIGN := CODE_SIGN_IDENTITY="-" DEVELOPMENT_TEAM="" CODE_SIGN_STYLE=Automatic
endif
```

## Workaround

```
make run DEV_SIGN='CODE_SIGN_IDENTITY="-" DEVELOPMENT_TEAM="" CODE_SIGN_STYLE=Automatic'
```

tarunlnmiit:
## Summary
- `grep -c` always prints a count (`0` on no match), never an empty string, so `ifeq (,$(shell security find-identity -v -p codesigning ... | grep -c "Developer ID Application"))` was never true — the ad-hoc signing fallback for contributors without a Developer ID cert never fired.
- Fixes #45.

## Fix
Compare the `grep -c` output against the literal `0` instead of empty string.

## Test plan
- [x] `security find-identity -v -p codesigning` shows `0 valid identities found` on this machine
- [x] Before fix: `make run` fails with `error: No signing certificate "Developer ID Application" found...`
- [x] After fix: plain `make run` (no manual `DEV_SIGN` override) succeeds and launches Codenotch.app

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ DEST    := platform=macOS,arch=arm64
 # changes nothing: project.yml's stable identity is what keeps a keychain
 # "Always Allow" grant alive across rebuilds, and forcing ad-hoc there would
 # throw that away and bring the prompt back on every `make run`.
-ifeq (,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
+ifeq (0,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
 DEV_SIGN := CODE_SIGN_IDENTITY="-" DEVELOPMENT_TEAM="" CODE_SIGN_STYLE=Automatic
 endif
 


vinzdg:

The ad-hoc signing fallback fires correctly now.

Genuinely appreciated.

PR #47, opened 2026-09-08
Add Scripts/sign-local.sh so local keychain grants persist

at10249:
Small contributor-DX helper, independent of #25.

### Problem
`make run`/`make build` produce an **ad-hoc-signed** app. An ad-hoc build has no stable code identity, so the macOS keychain access-control list can't match it to a saved **Always Allow** — and the prompt to read Claude Code's (or another tool's) token comes back on **every launch**. Since building locally is the only way to run a Debug build, anyone hacking on Codenotch hits this.

### Fix
`Scripts/sign-local.sh` creates a reusable **self-signed** `Codenotch Local Signing` identity in the login keychain (no Apple Developer account, no notarization — official releases are still Developer ID signed by `make release`) and re-signs the built app. The keychain grant then binds to a stable identity and sticks.

```sh
Scripts/sign-local.sh            # signs /Applications/Codenotch.app
Scripts/sign-local.sh path/to/Codenotch.app
```

Documented under **Building** in the README. Notable detail: OpenSSL 3 needs `-legacy` on the `pkcs12 -export` or `security import` fails with "MAC verification failed" — the script handles both.

### Testing
Verified on macOS 15: after signing, the keychain "Always Allow" for `Claude Code-credentials-<hash>` persists across relaunches (previously re-prompted every launch). Syntax-checked with `bash -n`. Happy to adjust naming/placement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,19 @@ certificate and an App Store Connect notary profile, and is only ever run by
 the maintainer to cut an official release. See
 [CONTRIBUTING.md](CONTRIBUTING.md).
 
+A Debug build is ad-hoc signed, which means it has no stable code identity, so
+macOS cannot match it to a saved keychain "Always Allow" — the prompt to read a
+tool's token returns on every launch. To make the grant stick during local
+development, sign the built app with a stable self-signed identity:
+
+```sh
+Scripts/sign-local.sh   # signs /Applications/Codenotch.app (pass a path to override)
+```
+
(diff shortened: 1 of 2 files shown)


vinzdg:
Still open — `Scripts/` only has `inspect-antigravity.sh`, so this hasn't landed. The keychain-grant problem it solves is real and documented in CONTRIBUTING; leaving it open.

PR #49, opened 2026-09-08
Add Simplified Chinese localization

DrSmoothl:
## Summary

Codenotch's visible copy was English-only. It now follows the Mac's language for English and Simplified Chinese.

English wording is unchanged and remains the source. Settings, menus, the notch tooltip, sign-in guidance, and What's New pick up Simplified Chinese when that is the system language. Brand names stay as they are.

## Test plan

- [x] `make test` — 563 tests, 0 failures, including new English / Simplified Chinese coverage for elapsed copy, reset copy, usage summaries, and menu strings
- [x] `make run` — Debug build launched; the bundle includes `en` and `zh-Hans`


vinzdg:
Still open — there's no localization infrastructure in the app at all yet (no `.lproj`, no `NSLocalizedString`), so this is a bigger change than a single language and I don't want to close it as though it were done. Leaving it open.

DrSmoothl:
Could I take the localization layer?

You're right it's more than one language. This PR already adds a String Catalog — the current stand-in for `.lproj` / `NSLocalizedString` — and routes visible copy through it, with English unchanged. I'd like to keep going here so adding a language is catalog + the localization list. Happy to split that into a follow-up if you'd rather.

This comment was translated from Simplified Chinese by Grok. Please forgive any ambiguity.

PR #50, opened 2026-09-08
Name a scoped weekly window after its model

HeeSJung:
## What

Claude's usage endpoint now returns a third limit of kind `weekly_scoped`, carrying a `scope` object (`scope.model.display_name`, e.g. "Fable", or a `surface`). Codenotch rendered it as a generic "Scoped" row, so a per-model weekly ring was indistinguishable from the account-wide one.

This decodes `Limit.scope`, labels the window by the scope's name ("Fable"), and gives it the id `weekly_scoped:<name>` so two scoped windows never collide. Untouched: `session` / `weekly_all` ordering and the headline id.

## Tests

Two new cases in `Tests/UsageResponseTests.swift`: a scoped window is named after its model, and an unscoped `weekly_scoped` still falls back to "Scoped".

Honest caveat: I could not run `make test` here (the CI-less build box signs ad-hoc over ssh and the XCTest bundle would not resolve the app module under those overrides). The change is built and running on my Mac as a Release ad-hoc build, where the tooltip shows the model name. If you have a local `make test`, that is the check to run.

## Why upstream

The scoped window is what a Max plan with a per-model cap shows every day; without this the ring is labelled by its kind, not by what it meters.

https://claude.ai/code/session_01D66oSpJGN2dn3Q4E4Bv88x

### DIFF
--- a/Sources/Providers/ClaudeOAuthProvider.swift
+++ b/Sources/Providers/ClaudeOAuthProvider.swift
@@ -241,6 +241,32 @@ struct UsageResponse: Decodable {
         let kind: String
         let percent: Double
         let resetsAt: Date?
+        /// Set on `weekly_scoped`: the model or surface the window covers. The
+        /// kind is the same for every such window; the scope carries the name.
+        let scope: Scope?
+
+        /// `weekly_scoped` appears once per scoped model, so the kind alone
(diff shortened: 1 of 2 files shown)


vinzdg:

A scoped weekly window is named after the model it meters.

Genuinely appreciated.

PR #51, opened 2026-09-08
Do not draw a ring for a directory with no token

### LINKED ISSUE #42: `~/.claude-mem` is detected as a Claude Code profile, and its row tells you to sign into it
## What happens

Codenotch shows a permanent extra ring, **Claude (mem)**, for an account that does not exist. Its row says:

> Sign in to Claude Code in `~/.claude-mem` to read your usage

`~/.claude-mem` is not a Claude Code profile. It is the data directory of the [claude-mem](https://github.com/thedotmack/claude-mem) plugin (93k stars), which stores its SQLite database, Chroma index, logs and worker state there. There is no account behind it and never will be.

## Why it is detected

`ClaudeProfile.discover` takes any `~/.claude-<slug>` directory, so `.claude-mem` becomes slug `mem`. `isProfileDirectory` then guards against exactly this — "a stray one someone made by hand would otherwise put a permanent sign-in ring in the notch" — but it accepts a directory if **any one** of the markers is present:

```swift
private static let markers = ["sessions", "projects", "settings.json",
                              "history.jsonl", ".claude.json"]
```

`settings.json` is the collision. claude-mem keeps its own `settings.json` there, so the guard passes on the one marker that is not distinctive. On my machine that is the only match:

```
~/.claude-mem
  settings.json      <-- matches (claude-mem's own config)
  claude-mem.db      chroma/      logs/      observer-sessions/
  supervisor.json    worker.pid   backups/   corpora/
```

Note `observer-sessions/`, not `sessions/` — the other four markers are all absent.

## Why the suggested fix is worth avoiding

The row's guidance is `CLAUDE_CONFIG_DIR=~/.claude-mem claude`. Following it makes Claude Code adopt another tool's working directory as its config: it writes `projects/`, `sessions/` and `.claude.json` into it, merges into the `settings.json` that belongs to claude-mem, and mints a second OAuth token in the keychain under the hashed service name. That is a real cost paid to silence an empty ring.

## Suggested fix

Dropping `settings.json` from `markers` looks sufficient and low-risk. On a real profile the other markers are there anyway — my `~/.claude` has `sessions`, `projects`, `history.jsonl` and `settings.json`, so it stays detected on three — while a foreign tool that happens to keep a `settings.json` no longer qualifies. `.claude.json` is the strongest signal for the non-default case specifically, since Claude Code writes it *inside* a custom `CLAUDE_CONFIG_DIR` (it is absent inside the default `~/.claude`, where it lives beside it as `~/.claude.json`).

A directory-name denylist would work too, but it needs updating for every new tool that borrows the `.claude-` prefix; tightening the marker fixes the class.

## Environment

- Codenotch 1.5.0 (from `hivinz.com`, signed and notarized)
- macOS 26.3 (25D125)
- claude-mem 13.3.0

ashiq54689:
ClaudeProfile.discover accepted any ~/.claude-<slug> directory that held one
of Claude Code's first-run filenames. claude-mem keeps its state in
~/.claude-mem and writes those same names, so it was discovered as an account
and drew a permanent "Sign in to Claude Code in ~/.claude-mem to read your
usage" ring — advice that cannot be followed, for a limit that does not exist.

Filenames cannot distinguish a plugin's data directory from a config
directory, and a denylist of plugin names would only postpone the next one.
Claude Code files a token under a service name derived from the config
directory's path, which ClaudeProfile already computes, so ask whether that
token exists: no token, no account, no ring.

The check is injected as hasCredential so the filename rules stay testable
without touching the login keychain. It tests existence, not validity — an
expired token still means the account is real, and the provider already
degrades that to credentialExpired and shows the last reading with its age.
The default profile is unaffected: ~/.claude is added unconditionally and
still gets the ring that says to sign in.

Closes #42

### DIFF
--- a/Sources/Providers/ClaudeProfile.swift
+++ b/Sources/Providers/ClaudeProfile.swift
@@ -37,21 +37,57 @@ struct ClaudeProfile: Equatable, Hashable {
     /// Code has actually used, slugs in alphabetical order so the rings never
     /// swap places between launches.
     ///
-    /// "Actually used" is judged by the files Claude Code writes on its first
-    /// run — an empty directory, or a stray one someone made by hand, would
-    /// otherwise put a permanent "sign in" ring in the notch for an account
-    /// that does not exist.
+    /// "Actually used" is judged twice over. The files Claude Code writes on
+    /// its first run rule out an empty directory or a stray one someone made
+    /// by hand; a token filed under the directory's own service name rules out
+    /// everything else that has learned to live at `~/.claude-<slug>`.
+    ///
+    /// The second test is what keeps plugins out. `claude-mem` keeps its state
+    /// in `~/.claude-mem` and writes every one of the first-run names above, so
+    /// the filename rules pass it and it is not an account: Claude Code has
+    /// never signed in there and never will, so the ring could only ever read
+    /// "Sign in to Claude Code in ~/.claude-mem to read your usage" — advice
+    /// that cannot be followed, for a limit that does not exist. Filenames
+    /// alone cannot tell the two apart, and a denylist of plugin names would
+    /// only postpone the next one. The credential can: no token, no account,
+    /// no ring.
+    ///
+    /// `hasCredential` is injected so discovery stays testable — the real one
+    /// reads the login keychain, which a test has no business touching. It only
+    /// enumerates attributes and takes a persistent reference, neither of which
+    /// needs authorization, so this costs no extra prompt per candidate.
     static func discover(home: URL = homeDirectory,
-                         fileManager: FileManager = .default) -> [ClaudeProfile] {
+                         fileManager: FileManager = .default,
+                         hasCredential: (ClaudeProfile) -> Bool = Self.hasKeychainCredential)
+    -> [ClaudeProfile] {
         let names = (try? fileManager.contentsOfDirectory(atPath: home.path)) ?? []
         let extras = names.compactMap { name -> ClaudeProfile? in
             guard let slug = slug(fromDirectoryName: name) else { return nil }
             let directory = home.appendingPathComponent(name)
             guard isProfileDirectory(directory, fileManager: fileManager) else { return nil }
-            return ClaudeProfile(slug: slug, configDirectory: directory)
+            let candidate = ClaudeProfile(slug: slug, configDirectory: directory)
+            guard hasCredential(candidate) else {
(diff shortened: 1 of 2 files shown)


ashiq54689:
The case that surfaced this — claude-mem keeps its state in `~/.claude-mem`, and that directory carries every first-run marker at once:

So `isProfileDirectory` passes it on any of the five names, and the notch showed a permanent "Sign in to Claude Code in ~/.claude-mem to read your usage" ring for an account that cannot exist. Gone after this change, with `~/.claude` unaffected.

Related: with the credential check in place I could sign two real accounts into `~/.claude-QD` and `~/.claude-ASF` and get correct per-account rings, which is what I was originally after.

vinzdg:
Still open — `ClaudeProfile.discover()` currently filters on marker files (`sessions`, `projects`, `settings.json`…) and not on whether a token actually exists, so a directory that was never signed in still gets a ring. The bug you describe is real and unfixed.

ashiq54689:
Rebased onto 1.6.0 and force-pushed. Green on your own `ci.yml` via `make test-ci`: https://github.com/ashiq54689/codenotch/actions/runs/34219622909

Two things changed from the original: it now asks across `keychainServices` rather than the primary name alone, so the either-spelling case you added for the default profile is respected here too, and the tests are redone against the current `ClaudeProfileTests`.

On your note — I think this may have been read as a bug report rather than a patch. The marker-only filtering you describe is exactly what it changes: `discover()` takes a `hasCredential` closure defaulting to a keychain lookup, and a candidate is returned only if both the marker check and that check pass. `isProfileDirectory` is untouched. `testADirectoryWithNoTokenIsNotAnAccount` pins both halves; `testTheDefaultSurvivesHavingNoToken` pins that `~/.claude` keeps its sign-in ring regardless.

The case that surfaced it, since the listing didn't render in my earlier comment — `claude-mem` keeps its state in `~/.claude-mem`, and on my machine that directory holds every first-run marker at once:

~~~
. .. .claude.json backups cache history.jsonl plugins projects sessions settings.json
~~~

So `isProfileDirectory` passes it on any of the five names, and the notch carried a permanent "Sign in to Claude Code in ~/.claude-mem to read your usage" ring for an account that cannot exist.

Happy to rework it your way if you would rather gate on something else — `signedInAddress()` reading `.claude.json` would distinguish a plugin directory just as well, and without a keychain call at all.

PR #52, opened 2026-09-08
Build on CI, and fix the ad-hoc signing guard that never fired

ashiq54689:
project.yml pins CODE_SIGN_IDENTITY to "Developer ID Application" with the
maintainer's team id and CODE_SIGNING_REQUIRED=YES, and the Makefile is meant
to override that with ad-hoc signing on any machine without that certificate.
The guard tested the wrong thing: `grep -c` writes a count to stdout whether
or not it matched, so $(shell ...) is "0" rather than empty and `ifeq (,...)`
never held. DEV_SIGN was empty everywhere, including on the maintainer's own
machine where it is supposed to be, so nobody was getting the fallback and a
contributor's first `make test` failed on a certificate they cannot have.
Compare against "0" instead.

With that working, .github/workflows/build.yml runs gen, test and build on a
macos-26 runner and uploads the ad-hoc signed app. That makes the project
buildable from a fork with no local Xcode, which is a 15GB install for anyone
who only wants to verify a small change. Pinned to macos-26 rather than
macos-latest because project.yml's deployment target needs the macOS 26 SDK,
and packaged with ditto because upload-artifact does not preserve the symlinks
inside a .app.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,56 @@
+# Builds and tests on a GitHub-hosted macOS runner, so a fork can be built
+# without a local Xcode install. The app is ad-hoc signed: the Makefile's
+# DEV_SIGN fallback fires because no Developer ID certificate exists on a
+# runner, which is the same path any contributor takes.
+name: build
+
+on:
+  push:
+    branches: [main, develop]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    # macos-26 carries the macOS 26 SDK that project.yml's deployment target
+    # needs. macos-latest points here too, but is pinned so a runner-image
+    # rollover cannot silently drop the SDK out from under the build.
+    runs-on: macos-26
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Toolchain
+        run: |
+          xcode-select -p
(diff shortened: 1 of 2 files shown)


ashiq54689:
Verified on a `macos-26` runner — Xcode 26.6, Swift 6.3.3, Sparkle 2.9.6, full suite green: https://github.com/ashiq54689/codenotch/actions/runs/34203443254

`DEV_SIGN` fires for the first time with this change; the invocation now reads `CODE_SIGN_IDENTITY="-" DEVELOPMENT_TEAM="" CODE_SIGN_STYLE=Automatic` and signing reports "Sign to Run Locally". Before it, `$(shell ... | grep -c ...)` evaluates to the string `0` rather than empty, so `ifeq (,...)` never held and xcodebuild fell through to project.yml's hardcoded Developer ID. CONTRIBUTING.md says none of the build targets need an Apple Developer account, and that wasn't reachable.

The workflow shares the commit because it can't go green without the Makefile fix. Happy to drop it and leave just the guard change if you'd rather not take CI from a contributor.

vinzdg:

CI builds and tests on every push, and the signing guard that never fired (`grep -c` returning "0" rather than nothing) is fixed.

Genuinely appreciated.

PR #53, opened 2026-09-08
Focus host application or terminal when clicking a live session row (#29)

### LINKED ISSUE #29: Click a live session to focus the app or terminal that owns it
## Problem

Codenotch can already tell me that an agent is waiting, name the live session, show where it is running, and say what it needs. The last step is still manual: I have to search through Terminal, iTerm, Cursor, VS Code, or Codex windows to find that session.

This is especially costly when several agents are running at once. The notch has already answered “which session needs me?” but cannot yet take me there.

`TASKS.md` already carries this as an unfinished M4b item: `Click a session to focus its terminal window`.

## Proposed behavior

- A live session row in the provider tooltip looks clickable when Codenotch can resolve where it came from.
- Clicking it activates the owning application and focuses the best matching window or tab.
- Waiting sessions remain first in the list, so the agent asking for input is the shortest click away.
- If an exact target cannot be resolved safely, the row stays informational rather than pretending the click will work.
- Clicking a session must not toggle the notch's pinned state or refresh provider usage.

An initial provider set could be Claude Code in Terminal/iTerm and sessions owned by Cursor, VS Code, or the Codex app. Other monitors can add focus support independently when they expose enough identity.

## Permission tradeoff

Bringing an application forward is straightforward, but choosing an exact terminal window or tab may require Apple Events or Accessibility permission depending on the host application. I would prefer this order:

1. Use an existing provider deep link or native activation API when available.
2. Activate a window already identified by the session monitor without broader access.
3. Request a targeted macOS permission only when the user first clicks a session that needs it, with clear copy explaining why.

No permission should be requested merely because Codenotch launched, and the feature should remain optional.

Would exact-window focus with that permission policy fit the project, and which host applications would you want in the first PR?

## Possible implementation seam

`AgentSession` could carry an optional focus target produced by each activity monitor. The tooltip would only add button behavior when that target exists, while a small focus coordinator would own application-specific activation. This keeps Terminal/iTerm/Cursor details out of the shared SwiftUI row.

## Acceptance checks

- Clicking a resolvable session focuses the intended application/window.
- Multiple sessions belonging to one application do not all focus the same arbitrary window.
- A dead or stale target fails quietly and does not activate the wrong session.
- Informational rows retain their current appearance and behavior when no target exists.
- Existing ring refresh, tooltip hover grace, and manual pinning behavior remain intact.

divingsbysangam:
Resolves #29

### Summary
Makes live session rows in the provider tooltip interactive and clickable when a focus target can be resolved, bringing the owning host application or terminal window to the front.

### Implementation Details
- **Focus Targets on `AgentSession`**: Introduced `AgentSession.FocusTarget` (`.process(pid_t)` and `.application(bundleID: String)`).
- **Target Resolution**:
  - `ClaudeSessionRecord`: Provides `.process(pid)`.
  - `GrokActivityMonitor`: Provides `.process(pid)` when present.
  - `CursorActivityMonitor`: Provides `.application(bundleID: CursorCredentials.bundleID)`.
  - `CodexActivityMonitor`: Provides `.application(bundleID: "com.openai.chat")` for desktop threads, while rollouts without a known host remain informational (`nil`).
- **`SessionFocus` Activator**:
  - For `.process(pid)`: Walks the process tree ancestors using `sysctl` (`KERN_PROC_PID` / `kinfo_proc.kp_eproc.e_ppid`) to find the nearest GUI ancestor application (Terminal, iTerm2, Warp, Ghostty, VS Code, Cursor) and activates it.
  - For `.application(bundleID)`: Finds running applications with the bundle identifier or executable name and activates them.
- **Event Handling & Notch Pinning**:
  - `NotchPanel.mouseDown`: Forwards `super.mouseDown(with: event)` so SwiftUI buttons inside the tooltip card receive click events.
  - `NotchWindowController.handleClick`: Restricts `togglePinned()` and provider refresh checks to `notchRect.contains(local)`, ensuring clicking within the tooltip card does not toggle the notch pinned state.
  - Restricted pointer hand cursor to notch cells and orb handle, avoiding whole-card cursor override and allowing clickable session rows to control their own hover state.
- **UI & Layout Integrity**:
  - Clickable session rows render using a plain button style with pointing hand cursor on hover and subtle detail text brightening.
  - Preserves exact layout line heights (`2 * cardBodyLineHeight + sessionRowGap`) so `NotchLayout.cardHeight` calculations and clipping bounds remain exact.
  - Unresolvable sessions remain informational rows without pointer cursor or click handlers.

### Verification
- Ran full test suite (`xcodebuild ... test`); all 562 unit tests passed.
- Added comprehensive unit tests in `SessionFocusTests` covering process tree ancestor resolution, bundle activation, and focus target assignments.

### DIFF
--- a/Sources/Features/TooltipCard.swift
+++ b/Sources/Features/TooltipCard.swift
@@ -335,6 +335,7 @@ private struct BlockedRow: View {
 private struct SessionRow: View {
     let session: AgentSession
     let now: Date
+    @State private var isHovered = false
 
     private var stateColor: Color {
         switch session.state {
@@ -360,7 +361,7 @@ private struct SessionRow: View {
         return session.detail
     }
 
-    var body: some View {
+    private var rowContent: some View {
         VStack(alignment: .leading, spacing: 0) {
             SplitRow(leading: session.name, trailing: stateWord,
                      trailingColor: stateColor) {
@@ -369,9 +370,38 @@ private struct SessionRow: View {
             SplitRow(
                 leading: detail,
                 trailing: ElapsedCopy.text(since: session.since, now: now),
-                leadingColor: Palette.textSecondary
+                leadingColor: (session.focusTarget != nil && isHovered) ? Palette.textPrimary : Palette.textSecondary,
+                trailingColor: (session.focusTarget != nil && isHovered) ? Palette.textPrimary : Palette.textSecondary
             )
             .padding(.top, NotchLayout.sessionRowGap)
+        }
+        .contentShape(Rectangle())
+    }
+
+    var body: some View {
+        if let target = session.focusTarget {
+            Button {
+                SessionFocus.activate(target: target)
+            } label: {
+                rowContent
+            }
+            .buttonStyle(.plain)
+            .onHover { inside in
+                isHovered = inside
+                if inside {
+                    NSCursor.pointingHand.push()
+                } else {
+                    NSCursor.pop()
+                }
+            }
+            .onDisappear {
+                if isHovered {
+                    isHovered = false
+                    NSCursor.pop()
+                }
+            }
+        } else {
+            rowContent
         }
     }
 }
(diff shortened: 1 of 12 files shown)


vinzdg:

Clicking a live session row in the tooltip now raises that session's application.

Genuinely appreciated.

PR #58, opened 2026-09-08
fix: fall back to ad-hoc signing without a valid development identity

RrezonCurraj:
A contributor can have an Apple Development certificate in their keychain while `security find-identity -v -p codesigning` reports zero valid identities. The Makefile extracted a team from that certificate and selected manual signing, causing local builds to fail because no usable certificate/private-key pair was available.

Read the development team from the first valid Apple Development signing identity instead, so this case falls back to ad-hoc signing. Existing Developer ID precedence is preserved.

Adds a shell regression check to CI covering the certificate-only case, valid development signing, multiple development identities, and Developer ID precedence across build/test/run dry runs.

Validation:
- Regression check fails against the original Makefile and passes with the fix.
- `make build` succeeds on the affected Mac; `codesign` confirms ad-hoc signing.
- `make test`: 753 tests passed, zero failures.
- `make run` succeeds and launches the local build.

Valid-certificate paths were checked with mocked identity output; the affected Mac has no valid signing identities.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,9 @@ jobs:
       - name: Install XcodeGen
         run: brew install xcodegen
 
+      - name: Check local signing detection
+        run: bash Scripts/test-signing.sh
+
       # Unsigned on purpose: no identity exists on a runner, and the tests
       # read fixtures rather than keychains — see `make test-ci`.
       - name: Run unit tests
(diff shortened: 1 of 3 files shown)


vinzdg:
Merged — thank you, and good catch on the root cause: `find-certificate` matches a certificate whose private key is gone, so the build was handed a `DEVELOPMENT_TEAM` it could never sign with. I checked `Scripts/test-signing.sh` against the old logic and it does fail there (`DEVELOPMENT_TEAM="ORPHAN1234"`), so the test earns its place rather than just documenting the new behaviour. It now runs in CI.

PR #59, opened 2026-09-08
fix: keep the settings button onscreen when dragging the notch

RrezonCurraj:
Dragging a left- or right-edge notch fully down lets the settings button extend below the screen, leaving it clipped and difficult to click. The drag clamp accounted for the notch body but treated the padding containing the settings handle as disposable hover-card space.

Reserve the settings handle’s full hit area beyond the body when clamping the panel position. The same trailing-end allowance applies to horizontal edges, where the handle can otherwise extend past the right edge. The default remains unchanged for callers without a trailing extent.

Validation:
- `make test`: 754 tests passed, zero failures.
- Added a regression test covering all four edges and a screen with a nonzero origin.
- Local build succeeded; the reporter confirmed the updated app fixes the clipping after manually testing it.

The local test/build run included the signing workaround in #58. This PR contains only the settings-handle geometry fix and its test.

### DIFF
--- a/Sources/Notch/NotchGeometry.swift
+++ b/Sources/Notch/NotchGeometry.swift
@@ -90,7 +90,10 @@ enum NotchGeometry {
         // instead — `panelSize` shrunk by this on each end — lets it travel
         // almost the full edge; the padding is free to run past the bezel,
         // since nothing is drawn there until a card actually opens.
-        slack: CGFloat = 0
+        slack: CGFloat = 0,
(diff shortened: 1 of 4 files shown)


vinzdg:
Merged — thank you. Nudging the pill to the end of its travel could push the orb off the bezel, and with the notch hidden that is the only visible way into Settings — so this was a genuine way to get locked out. Reserving `trailingExtent` on all four edges is the right shape for the fix.

PR #60, opened 2026-09-08
Support multiple Codex accounts with isolated profiles

Const-Iv:
Codex currently reads only `~/.codex`, so a second account used through `CODEX_HOME=~/.codex-work` has no separate usage ring or activity indicator. This adds the same profile discovery convention already used for Claude: the default account stays `Codex`, while used `~/.codex-<slug>` directories appear as `Codex (slug)` after restarting Codenotch.

Each profile reads its own credentials and activity databases, has an independent retry deadline, and can be reordered or disconnected in Settings without affecting another profile. Existing default-account IDs and preferences are preserved. The README documents signing into additional accounts through Codex CLI; named profiles also show the corresponding sign-in command. Codenotch only reads credentials and leaves sign-in and token refresh to Codex.

Validation:
- `make test` on macOS 26.6.2 / Xcode 26.6: 764 tests passed.
- Added coverage for discovery and ordering, shell quoting, per-profile credentials and requests, token rotation, retry isolation, disconnect behavior, and separate CLI/desktop activity databases.
- Local installed fork successfully reads live Codex usage; the user also confirmed the additional-account setup works. Automated tests use synthetic credentials and intercepted requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,31 @@ Any `~/.claude-<slug>` directory Claude Code has run against is found at launch;
 the default `~/.claude` always comes first, the rest in alphabetical order, so the
 rings never swap places.
 
+Codex accounts work the same way: `~/.codex` stays the **Codex** ring, and each
+used `~/.codex-<slug>` directory adds a **Codex (slug)** ring with its own limits,
+activity and Settings row. Profiles are discovered at launch, default first,
+then alphabetically. To connect a second account, sign in through Codex CLI
+using a separate home directory:
+
+```sh
+mkdir -p "$HOME/.codex-work"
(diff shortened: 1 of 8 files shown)


gusttavocabral:
Thanks for adding isolated profiles for multiple Codex accounts. A related UX request: could Codenotch eventually offer an in-app way to add and manage multiple Codex and Claude accounts? The current CODEX_HOME / CLAUDE_CONFIG_DIR setup is useful, but it requires CLI commands and restarting the app. It would be great to sign in to a second account directly from Settings, show each account as its own ring, and manage re-authentication or enable/disable state independently.

PR #64, opened 2026-09-08
Support macOS Sequoia (15.0)

nicolasembleton:
Lower the deployment target from macOS 26.0 to 15.0 so the app runs on Sequoia.

The only macOS 26-only API is `glassEffect` in `SettingsView`, used for the sidebar and collapse button backgrounds. Gated behind `#available(macOS 26.0, *)` with a `.regularMaterial` fallback — the visual difference is minor and all interactions are unchanged.

Also adds `make install`: builds a Release `.app` with the same auto-detected signing identity (Developer ID, Apple Development, or ad-hoc) and copies it to `/Applications`. Useful for contributors who want a permanent copy without the notarized release path.

All 753 tests pass against the 15.0 target on macOS 15.8.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ DEV_SIGN := CODE_SIGN_IDENTITY="Apple Development" CODE_SIGN_STYLE=Manual \
 endif
 endif
 
-.PHONY: gen build test test-ci run clean
+.PHONY: gen build test test-ci run install clean
 
 gen:
 	xcodegen generate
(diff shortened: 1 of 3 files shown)


vinzdg:
Thank you for this — and sorry to close it rather than merge it.

Two PRs arrived within hours doing the same thing: this one and #65, which additionally parameterises the architecture so Intel (x86_64) builds too. They are mutually exclusive — both rewrite the deployment target — so I took the broader one, and #65 is now merged.

Your diagnosis was identical and correct: the floor was macOS 26 purely because the settings panel uses Liquid Glass, and guarding `glassEffect` behind `#available` with a material fallback is exactly the right fix. That is what shipped, via #65.

Genuinely appreciated — this was a real limitation and you were among the first to name it.

nicolasembleton:
Thanks for the update and no worries at all. If the other one is better, then I'm glad it's merged in :) I selfishly just needed Sequoia and didn't even think of adding Intel-based archs which was a miss in hindsight 😅. Thanks for the word and the app :)

nicolasembleton:
BTW @vinzdg will Sparkle also support the lesser targets? It's based on the project configuration right? I had issues with the update process due to that, and I did not check on my fork, so raising this in case it requires attention.

Issue #17, opened 2026-09-06
Add support for Grok Build


0xBush:

aaronjmars:
Concrete payload for a Grok Build Plan account, in case it helps implement this.

`GrokLocalProvider` gets HTTP 200 from `cli-chat-proxy.grok.com/v1/billing?format=credits`, but a Build Plan account with `isUnifiedBillingUser: true` returns neither `creditUsagePercent` nor `productUsage`, so `GrokUsage.windows(...)` throws `nothingMetered` and the cell is blank. Every leaf of the response:

```
config.currentPeriod.type      = USAGE_PERIOD_TYPE_WEEKLY
config.currentPeriod.start     = 2026-09-07T20:59:12+00:00
config.currentPeriod.end       = 2026-09-14T20:59:12+00:00
config.onDemandCap.val         = 0
config.onDemandUsed.val        = 0
config.isUnifiedBillingUser    = true
config.prepaidBalance.val      = 0
config.topUpMethod             = TOP_UP_METHOD_SAVED_PAYMENT_METHOD
config.billingPeriodStart      = 2026-09-07T20:59:12+00:00
config.billingPeriodEnd        = 2026-09-14T20:59:12+00:00
```

`?format=usage` adds a numerator but no denominator:

```
config.monthlyLimit.val = 0
config.used.val         = 14
config.onDemandCap.val  = 0
```

So there is a weekly window and a `used` value, but no allowance total and no percent on this billing token. Other paths (`/v1/usage`, `/v1/rate-limits`, `/v1/limits`, `/v1/quota`, `/v1/me`, `/v1/subscription`) all 404, and the grok CLI only ever calls `cli-chat-proxy.grok.com/v1`. `grok models` reports `agentType: grok-build-plan` and emits a `grok-shell-process_resource_limits` telemetry event, so the weekly limit is computed inside the `grok-shell` ACP subprocess rather than exposed on the billing endpoint.

If there is no public endpoint that returns the weekly Build Plan allowance, one option is to surface `used` with an explicit "no metered limit" state for these accounts instead of a blank cell.

Env: codenotch 1.6.0, macOS (Darwin 25.6.0), grok CLI 1.0.13.

aaronjmars:
[...]

PR #75, opened 2026-09-08
Show Grok weekly pool ring when currentPeriod has no percent yet

aaronjmars:
Draws the Grok weekly ring for accounts whose weekly pool is stated in `currentPeriod` with no `creditUsagePercent`/`productUsage` yet (X Premium+ / SuperGrok). Addresses #17.

## Problem

`GrokLocalProvider` reads `~/.grok/auth.json` and gets HTTP 200 from `cli-chat-proxy.grok.com/v1/billing?format=credits`, but for an X Premium+ weekly pool the response carries a weekly `currentPeriod` window and no `creditUsagePercent` and no `productUsage` until usage lands. `GrokUsage.windows(...)` then finds no window and throws `nothingMetered`, so the cell is blank.

Grok's own `/usage` shows this exact account as:

```
Weekly limit (X Premium+)
[                    ] 0%
Resets: September 14, 16:59
```

i.e. a 0% weekly bar with the period end as the reset. Codenotch had the same response in hand but treated a fresh (0%) period as an unmetered account.

Observed response for such an account:

```
config.currentPeriod.type   = USAGE_PERIOD_TYPE_WEEKLY
config.currentPeriod.end    = 2026-09-14T20:59:12+00:00   (= Sep 14 16:59 local, matches /usage)
config.onDemandCap.val      = 0
config.onDemandUsed.val     = 0
config.isUnifiedBillingUser = true
config.prepaidBalance.val   = 0
```

No `creditUsagePercent`, no `productUsage`.

## Change

`GrokUsage.windows(...)`: after the existing `creditUsagePercent` and `productUsage` branches produce nothing, if a weekly `currentPeriod` is present, emit a `credits` window at `usedFraction: 0` labelled `Weekly limit`, resetting at the period end. The existing `nothingMetered` guard still fires for a genuinely empty config (no `currentPeriod`), so `testAnEmptyConfigIsNotASuccessfulReading` is unchanged.

This mirrors what grok itself renders, rather than reaching for the weekly-pool total through a signed-in WebView (the approach in the closed #10). When usage lands and `creditUsagePercent` appears, the existing branch takes over and the ring fills.

## Tests

Added `testWeeklyPoolWithoutAPercentIsAZeroRing`: a weekly-`currentPeriod`, no-percent config yields one `credits` window, `label == "Weekly limit"`, `usedFraction == 0`, reset on Sep 14. Existing tests (percent ring, product-only ring, empty-config throws, garbage throws) unchanged.

### DIFF
--- a/Sources/Providers/GrokUsage.swift
+++ b/Sources/Providers/GrokUsage.swift
@@ -54,6 +54,23 @@ enum GrokUsage {
             }
         }
 
+        // A weekly plan pool (X Premium+, SuperGrok) states its window in
+        // `currentPeriod` and omits `creditUsagePercent`/`productUsage` until
+        // usage lands, so the branches above find nothing on a fresh period.
+        // Grok's own `/usage` still shows this as a "Weekly limit" bar at 0%
+        // with the period end as the reset, so mirror it rather than reporting
+        // the account as unmetered.
+        if windows.isEmpty,
+           let weekly = period(credits["currentPeriod"]),
+           (weekly["type"] as? String).map({ $0.contains("WEEKLY") }) == true {
+            windows.append(LimitWindow(
+                id: "credits",
+                label: "Weekly limit",
+                usedFraction: 0,
+                resetsAt: date(weekly["end"]) ?? creditsReset
+            ))
+        }
+
         guard !windows.isEmpty else {
             throw UsageProviderError.nothingMetered("Grok has nothing metered on this account yet")
         }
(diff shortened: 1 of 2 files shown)


vinzdg:
Merged — thank you. This one mattered: a fresh weekly period was being reported as an *unmetered account* rather than a 0% reading, which is the wrong answer to "where do I stand". Mirroring what Grok's own `/usage` shows is exactly right.

PR #84, opened 2026-09-09
fix: treat the Antigravity CLI as a real install

### LINKED ISSUE #82: Antigravity request count is always 0 on CLI-only installs: transcript root never checks ~/.gemini/antigravity-cli/brain
## Summary

The Antigravity "Requests today" count is always 0 for **CLI-only** Antigravity installations, because the transcript-root probe never checks the directory the CLI actually writes to.

## Details

`AntigravityActivity.transcriptRoot` (`Sources/Providers/AntigravityActivity.swift`, lines 19–27, current `main` @ `6f0eb97`) probes exactly two locations:

1. `~/.gemini/antigravity-ide/brain` (IDE layout)
2. `~/.gemini/antigravity/brain` (fallback)

An Antigravity **CLI-only** install uses neither. It writes trajectories to:

```
~/.gemini/antigravity-cli/brain/<trajectory>/.system_generated/logs/transcript.jsonl
```

On such a machine both probes fail, `~/.gemini/antigravity/brain` does not exist, `contentsOfDirectory` throws, and `read()` returns `requestsToday: 0, lastRequest: nil` — even with same-day CLI activity on disk.

Because `AntigravityActivityMonitor` is initialized with the same `transcriptRoot` default (`Sources/Sessions/AntigravityActivityMonitor.swift`, line 25), live-session detection ("Working" state) is also blind on CLI-only installs — it can never see any session.

## Steps to reproduce

1. Install the Antigravity CLI only (no IDE app).
2. Run a few agent requests today. Trajectories and `transcript.jsonl` files appear under `~/.gemini/antigravity-cli/brain/`.
3. Open the CodeNotch Antigravity cell (consumer-account / CLI-only path, where the derived local count is used).
4. The cell reports "no requests today" / 0 despite same-day activity.

## Expected behavior

The count should include trajectories under `~/.gemini/antigravity-cli/brain`.

## Actual behavior

Always 0 on CLI-only installs; `lastRequest` is never set; the session monitor never reports activity for the CLI.

## Notes (context, not bugs)

- The metric is inherently *requests today*: a user whose last Antigravity use was on a previous day will correctly see 0 even after the path fix.
- `AntigravityActivity.lastRequest` is computed and stored but never surfaced anywhere in the UI (it is only referenced in tests). Exposing it — e.g. in the tooltip, the way other providers show recency — would help disambiguate "0 today" from "not used recently". Happy to have this split into its own issue if preferred.

## Suggested fix

Extend the probe chain, e.g. IDE → CLI → legacy:

```swift
static var transcriptRoot: URL {
    let home = NSHomeDirectory()
    for candidate in ["antigravity-ide", "antigravity-cli", "antigravity"] {
        let dir = URL(fileURLWithPath: home)
            .appendingPathComponent(".gemini/\(candidate)/brain")
        if FileManager.default.fileExists(atPath: dir.path) { return dir }
    }
    return URL(fileURLWithPath: home)
        .appendingPathComponent(".gemini/antigravity/brain")
}
```

One nuance for the maintainer: if both the IDE and CLI directories exist and are actively used, "first existing wins" can undercount. Reading all existing roots and merging the counts would be more robust — the probe chain above is just the minimal change.

## Environment

- CodeNotch: current `main` (`6f0eb97`, 2026-09-09) — the code path above is unchanged there
- Antigravity: CLI-only install, consumer account (derived local count path)

mulhamna:
Closes #82.

## Summary

Antigravity ships a CLI as well as an IDE, and Codenotch only ever looked for the IDE. On a CLI-only machine that costs three things: the request count is always 0, the ring never spins while an agent is working, and — the one I did not expect — a **real quota is being served on loopback the whole time and never asked for**.

## One correction to #82 before the rest

The fix I suggested in the issue was a probe chain, `antigravity-ide` → `antigravity-cli` → `antigravity`, first one that exists wins. **It would not have fixed my own machine.** Moving between flavours does not remove the old directory, so all four exist:

| directory | `brain` exists | trajectories |
|---|---|---|
| `~/.gemini/antigravity` | yes | 0 |
| `~/.gemini/antigravity-ide` | yes | **0** ← the chain stops here |
| `~/.gemini/antigravity-cli` | yes | **37** |
| `~/.gemini/antigravity-backup` | yes | 0 |

"The first that exists" is not a choice between a real install and a missing one. It picks an empty directory while the transcripts sit in the next. So this reads **all** of them and adds them up — trajectories are UUID-named per install, so nothing is double counted, and somebody who uses the IDE and the CLI in one day gets one number for the day rather than whichever half was looked at first.

## Changes

**The activity count and the session monitor** — `AntigravityActivity.transcriptRoots` returns every `~/.gemini/antigravity*/brain` that exists, and `read` sums across them. `AntigravityActivityMonitor` watches all of them and reports the most recent turn, so the working state is no longer blind on a CLI-only install.

**The real quota** — `agy` serves the same `RetrieveUserQuotaSummary` on loopback that the IDE's language server does: two ports, the same response, the same shape `AntigravityBridge.windows(in:)` already parses without a change. Discovery only matched `language_server --csrf_token`, so the CLI was never found. It is now, and the provider reports `.official` percentages instead of a derived count.

The CLI asks for no token — on loopback it answers anyone — so the header is sent only where there is one to send: an empty header is not the same request as no header, and the IDE still refuses without it. The IDE therefore keeps first place when both are running. `agy` is matched on the executable's own name rather than anywhere in the process line, because three letters turn up inside real words and real paths and `/opt/legacy/bin/server` is not an Antigravity install.

**Recency** — the issue's closing note, included here rather than split off: `lastRequest` was already computed and thrown away. A bare `0` on a quiet day reads as the app having failed to find anything, which is exactly what the wrong directory looked like — part of why this went unnoticed. The row now says when it was last used. Counted in calendar days, because the number beside it is: `requestsToday` asks whether a timestamp falls on today's date, and measuring elapsed hours instead calls a Saturday evening "2 days ago" on a Tuesday morning.

## Test plan

- [x] `make test` passes; no new warnings.
- [x] The count, against real data: 213, 187 and 177 on the three busiest days, matching an independent count over the same files. Every one of them read 0 before.
- [x] Live, with `agy` running: the count moved 0 → 4 as requests were made, which is the only evidence that actually settles the directory question.
- [x] The quota, live through `AntigravityBridge` on a running `agy`: four windows — a weekly and a five-hour limit for the Gemini models, another pair for the Claude and GPT ones — with groups and reset times, and the provider reporting them as `.official`.
- [x] Both fixes fail their tests when reverted: the discovery test finds 1 root instead of 4, and the CLI endpoint comes back nil.
- [ ] **Not tried with the IDE installed**, so the "IDE wins when both are running" ordering is covered by a test over a fake process table and not by a real one. I do not have the IDE on this machine.
- [ ] Not tried on a machine where only `~/.gemini/antigravity` exists — the legacy layout — beyond the tests.

One thing worth flagging rather than leaving to be discovered: the quota is only there while `agy` is running. When it exits the bridge goes with it and the provider falls back to the counted requests, which is the existing `everBridged` behaviour and reads as a dimmed, dated last figure rather than a sudden change of units.

## Evidance
WhatsApp Image 2026-09-09 at 11 47 28

### DIFF
--- a/Sources/Providers/AntigravityActivity.swift
+++ b/Sources/Providers/AntigravityActivity.swift
@@ -16,14 +16,40 @@ struct AntigravityActivity: Equatable {
     let requestsToday: Int
     let lastRequest: Date?
 
-    static var transcriptRoot: URL {
-        let idePath = URL(fileURLWithPath: NSHomeDirectory())
-            .appendingPathComponent(".gemini/antigravity-ide/brain")
-        if FileManager.default.fileExists(atPath: idePath.path) {
-            return idePath
+    /// Every Antigravity install's transcripts, not just the first one found.
+    ///
+    /// Antigravity keeps a directory per flavour under `~/.gemini` —
+    /// `antigravity`, `antigravity-ide`, `antigravity-cli`, `antigravity-backup`
+    /// — and each has its own `brain`. Picking the first that *exists* looked
+    /// reasonable and was not: leaving a flavour behind leaves its directory
+    /// behind too, so on a machine that has run the IDE and then moved to the
+    /// CLI all four exist and the first is empty. The count came out zero while
+    /// the transcripts sat one directory over.
+    ///
+    /// So: all of them. Trajectories are UUID-named per install, so nothing is
+    /// counted twice, and somebody who uses the IDE and the CLI in the same day
+    /// gets one number for the day rather than whichever half was looked at.
+    static var transcriptRoots: [URL] { transcriptRoots(home: URL(fileURLWithPath: NSHomeDirectory())) }
+
+    static func transcriptRoots(home: URL) -> [URL] {
+        let gemini = home.appendingPathComponent(".gemini")
+        let names = (try? FileManager.default.contentsOfDirectory(atPath: gemini.path)) ?? []
+        return names
+            .filter { $0.hasPrefix("antigravity") }
+            .sorted()
+            .map { gemini.appendingPathComponent($0).appendingPathComponent("brain") }
+            .filter { FileManager.default.fileExists(atPath: $0.path) }
+    }
+
+    /// Every install's activity, as one day's worth.
+    static func read(roots: [URL] = transcriptRoots, now: Date = Date()) -> AntigravityActivity {
+        roots.reduce(AntigravityActivity(requestsToday: 0, lastRequest: nil)) { total, root in
+            let one = read(root: root, now: now)
+            return AntigravityActivity(
+                requestsToday: total.requestsToday + one.requestsToday,
+                lastRequest: [total.lastRequest, one.lastRequest].compactMap { $0 }.max()
+            )
         }
-        return URL(fileURLWithPath: NSHomeDirectory())
-            .appendingPathComponent(".gemini/antigravity/brain")
     }
 
     /// A step the model actually answered. User input and system checkpoints
@@ -31,7 +57,8 @@ struct AntigravityActivity: Equatable {
     /// work the model never did.
     private static let modelSource = "MODEL"
 
-    static func read(root: URL = transcriptRoot, now: Date = Date()) -> AntigravityActivity {
+    /// One install's, which is what the combining read above is made of.
+    static func read(root: URL, now: Date = Date()) -> AntigravityActivity {
         let manager = FileManager.default
         guard let trajectories = try? manager.contentsOfDirectory(
             at: root, includingPropertiesForKeys: nil
--- a/Sources/Providers/AntigravityActivity.swift
+++ b/Sources/Providers/AntigravityActivity.swift
@@ -112,4 +112,43 @@ struct AntigravityActivity: Equatable {
         guard requestsToday > 0 else { return "no requests today" }
         return "~\(requestsToday) request\(requestsToday == 1 ? "" : "s") today"
     }
+
+    /// What the tooltip's row is called.
+    ///
+    /// A bare `0` on a day Antigravity has not been opened reads as the app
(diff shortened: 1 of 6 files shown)


mulhamna:
@vinzdg — if it fits your plans, I would love to see this land in 1.7.0. Entirely your call on timing and on whether you want it in this shape at all.

The reason I am asking rather than just leaving it in the queue: on a CLI-only machine the Antigravity cell is not slightly wrong, it is blank. The count is 0 whatever you did today, the ring never spins while an agent is working, and the real quota — which turns out to be served on loopback the whole time — is never asked for. Someone in that position has no way to tell the difference between "Codenotch cannot read this" and "I have not used it", which is precisely what kept me from noticing the bug for as long as I did.

Two things that might make it easier to take:

- The quota half is additive. Discovery gains a second branch and the response parser is untouched — `AntigravityBridge.windows(in:)` already reads the CLI's response as-is, which is what convinced me the two servers really are the same one.
- If the recency line reads as scope creep, it is one commit (`54df08e`) and drops out cleanly. It is only there because a bare `0` and a broken path look identical from the outside.

Happy to split this into separate PRs, rebase once 1.7.0 is cut, or adjust anything that does not match how you would have written it — you rewrote #23 rather than merging it and the result was better than what I sent, so I would rather hear that early than have you work around it.

The two unchecked boxes in the description are real, not modesty: I do not have the Antigravity IDE installed, so "the IDE wins when both are running" is covered by a test over a fake process table and never by a real one. That is the part I would want a second pair of eyes on.

PR #85, opened 2026-09-09
Build a downloadable app bundle in CI, so trying Codenotch needs no Xcode

facundofarias:
## Why

Trying Codenotch means compiling it. `make release` needs the Developer ID
certificate and the stored notarization credentials, so it only ever runs on
your machine, and `make install` needs Xcode and a checkout. Between a shipped
release and a contributor there is nobody — someone merely curious has to
install a multi-gigabyte toolchain before they can see a single ring.

This adds a compiled build anyone can download and drag to Applications. No
Xcode, no checkout, no Apple account, no `make`. That matters more than usual
for this app: its whole pitch is that it reads tools you already have, so the
distance between "sounds useful" and "I can see my own numbers" ought to be one
download — and it currently is not.

It also gives contributors and you something to hand a bug reporter: a build of
`main` from before and after a change, without asking them to compile anything.

## What is here

Three commits, separable if you only want part of it:

1. **`make dmg-ci` and the Package workflow** — the same Release configuration,
   ad-hoc signed, packaged as a drag-to-Applications dmg and uploaded as an
   artifact on every push, every pull request, and on demand from any branch.
2. **`actions/checkout@v4` → `v7`** — every run was annotated for Node 20,
   which the runners already force onto Node 24 and will stop carrying.
3. **A rolling `preview` prerelease, and a README `## Download` section.**

## This changes what the repository publishes — worth your attention

Commit 3 means that once merged, a push to `main` publishes a `preview`
prerelease carrying that unsigned build. Constraints I worked to:

- **Its own tag, never `v*`.** Those belong to `make publish`, and an unsigned
  dmg attached beside a notarized one is how somebody ends up running the wrong
  download and clearing quarantine flags they never needed to.
- **Marked a prerelease**, so the "Latest" badge stays on the real release.
- **Refreshed in place**, not deleted and recreated, so watchers are notified
  once rather than on every merge. The tag is moved to the built commit so the
  release never advertises one commit while serving another.

Drop that commit and the first two still stand on their own — you would just
have artifacts rather than a link the README can point a stranger at.

## The non-obvious part

An ad-hoc build cannot load Sparkle without help, for the reason `make
install`'s comment already records. The hardened runtime turns on library
validation, which loads a library only when its Team ID matches the process's,
and an ad-hoc signature has no Team ID at all — so the app and the framework
beside it can never be shown to match. Both being ad-hoc is the problem, not
the escape from it.

Nothing catches this before launch. Each signature is individually valid, so
the build succeeds and `codesign --verify --deep --strict` reports the bundle
valid and satisfying its Designated Requirement. It is dyld that refuses:

```
Library not loaded: @rpath/Sparkle.framework/Versions/B/Sparkle
... not valid for use in process: mapping process and mapped file
(non-platform) have different Team IDs
```

which macOS shows the user as "Codenotch cannot be opened because of a
problem". `com.apple.security.cs.disable-library-validation` relaxes that one
check and nothing else, so a CI build still runs under the hardened runtime.

Separately, Xcode adds `com.apple.security.get-task-allow` to any
non-distribution signature, which lets another process attach to and read the
memory of an app holding other tools' OAuth tokens. `-exportArchive` would drop
it, but that is the step needing the Developer ID identity, so the signature is
replaced after the build instead, and the target fails if the entitlement comes
back.

## Verified, not assumed

I got this wrong once before catching it, so: every claim below was checked
against a build published by this workflow, not by reading the diff.

- Downloaded the artifact, mounted it, and **launched the app** — it runs, and
  `vmmap` confirms the framework dyld previously refused is mapped into the
  process.
- `codesign -d --entitlements` on the shipped bundle shows
  `disable-library-validation` alone; `get-task-allow` is gone.
- Signature flags still `adhoc,runtime` — the hardened runtime is intact.
- Both branches of the release step exercised: creating it, and refreshing it
  (tag moved, stale asset removed, `publishedAt` unchanged, so no second
  notification).
- The release step is correctly skipped on non-`main` refs.

## Left for you

- The README badge reads `macOS 26+` while `project.yml` now says `15.0`. My
  section says macOS 15 to match `LSMinimumSystemVersion`, which puts two
  different numbers in one README. One line to fix, but it is your call which
  is right — say the word and I will add it.
- Preview builds keep the shipped `SUFeedURL`, so like any build they check the
  appcast daily and a later release will replace them. Sensible for someone
  just trying the app, surprising for someone testing their own branch. Happy
  to disable automatic checks for CI builds if you would rather.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     # deployment target or the Liquid Glass APIs the settings panel uses.
     runs-on: macos-26
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
 
       # Printed rather than assumed: when this job fails on a runner image
       # change, the toolchain is the first thing worth seeing.
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -73,6 +73,8 @@ jobs:
             echo "xattr -dr com.apple.quarantine /Applications/Codenotch.app"
             echo '```'
             echo
+            echo "If macOS says the app is *damaged*, that is the quarantine flag rather than a bad download — run the command above."
+            echo
             echo "Universal binary — Apple Silicon and Intel. Requires macOS 15 or later."
           } >> "$GITHUB_STEP_SUMMARY"
 
@@ -109,6 +111,8 @@ jobs:
           Applications, clear the flag once:
 
               xattr -dr com.apple.quarantine /Applications/Codenotch.app
+
+          If macOS says the app is *damaged*, that is the quarantine flag rather than a bad download — run the command above.
 
           For a signed and notarized build, take the latest release instead.
 
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -14,6 +14,10 @@ on:
 jobs:
   package:
     name: Unsigned app bundle
+    # Only the preview release step writes anything; the token is read-only by
+    # default on this repository, and `gh release create` fails without this.
+    permissions:
+      contents: write
     # Same reason as the test job: an older image ships an Xcode that cannot
     # read the project at all — `objectVersion 77` is a format it does not
     # know — whatever the deployment target happens to be.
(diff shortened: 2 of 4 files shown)


mulhamna:
Hey, thanks for this — being able to try Codenotch without installing Xcode is exactly what I was hoping for 🙏

One thing that might trip people up: the `xattr` step. On newer macOS, a quarantined app often opens with **"Codenotch.app is damaged and can't be opened"** — which sounds fatal, has no right-click → Open escape hatch, and nothing points you at the `xattr` fix. I only knew what to do because I read the README all the way down.

If you're open to it, a tiny install script would make this dead simple — fetch the preview dmg, mount, copy to /Applications, clear quarantine, launch:

```sh
#!/bin/sh
set -eu

REPO="vinzdg/codenotch"
APP="/Applications/Codenotch.app"

url=$(curl -fsSL "https://api.github.com/repos/$REPO/releases/tags/preview" \
  | python3 -c 'import json,sys; a=json.load(sys.stdin)["assets"]; print(next(x["browser_download_url"] for x in a if x["name"].endswith(".dmg")))')

TMP=$(mktemp -d)
curl -fL "$url" -o "$TMP/codenotch.dmg"
mkdir -p "$TMP/mnt"
hdiutil attach -nobrowse -readonly -mountpoint "$TMP/mnt" "$TMP/codenotch.dmg"
rm -rf "$APP"
cp -R "$TMP/mnt/Codenotch.app" /Applications/
hdiutil detach "$TMP/mnt"

printf 'Clear the quarantine flag on %s? [y/N] ' "$APP"
read -r answer
case "$answer" in y|Y) xattr -dr com.apple.quarantine "$APP"; open "$APP";; *) echo "Skipped. Run it yourself later:"; echo "  xattr -dr com.apple.quarantine $APP";; esac
```

so the whole thing becomes one line for users:

```sh
curl -fsSL https://raw.githubusercontent.com/vinzdg/codenotch/main/scripts/install.sh | sh
```

Two small thoughts, feel free to ignore:

- I made the `xattr` step ask first instead of just doing it — I noticed you were already careful about people clearing quarantine flags they never needed to (the preview/`v*` split 👍), so a silent `xattr` on every install seemed against the spirit of that. Also keeping the path hardcoded to `/Applications/Codenotch.app` and never taking an argument, so nothing else can ever get recursed over.
- If a whole script feels like too much for this PR, even one line in the README / step summary would already help: *"If macOS says the app is damaged, that's the quarantine flag — run the `xattr` command above."*

Happy to test if you end up adding the script. Thanks again!

facundofarias:
Thanks — the "damaged" dialog is the right thing to have spotted, and it's the failure mode most likely to make someone give up. The instruction was already in the README; what was missing was the *symptom* that sends people looking for it, since "damaged" reads as a corrupt download rather than a flag to clear. Named now in all three places somebody meets the build — README, run step summary, and the preview release notes — in 4cf6e91.

On the installer I'd rather not, and I think your own two design points are most of the argument. Concretely:

1. **It breaks for exactly the people it's for.** `/usr/bin/python3` is `com.apple.dt.xcode_select.tool-shim-public`, not Python — on a Mac without Command Line Tools it prompts to install them. Someone with no Xcode is the entire audience for these builds, and the first thing the script would do is open a toolchain-install dialog.

2. **`rm -rf /Applications/Codenotch.app` with no confirmation** replaces a notarized, self-updating install with an unsigned one. Worse than it sounds: the code identity changes with it, so the keychain "Always Allow" grants are lost and the prompt to read each tool's token comes back on every launch. That's what `project.yml`'s stable-identity note is guarding.

3. **Piping unpinned code from `main` into a shell** to install the app that reads your OAuth tokens is a habit I'd rather not teach this particular user base — and nothing verifies the dmg it downloads, by checksum or signature.

Your points about making the `xattr` step explicit and scoping it to a fixed path are both right, and the fact that the safe version needs an interactive prompt seems to me an argument that a one-line `curl | sh` isn't the shape this wants.

If vinz wants an installer I'm glad to help with one that needs no Python, doesn't clobber an existing install without saying so, and verifies what it downloaded — but it's his distribution story, and I'd rather keep this PR to producing the build.

PR #102, opened 2026-09-09
feat: draw the weekly limit as a second ring

mulhamna:
## Summary

A cell draws one window — the headline the provider declares, which for Claude is the current session. The weekly allowance is the one that actually runs out first on a busy week, and the only way to see it was to hover. This gives it a ring of its own.

**Off by default**, so an existing install looks exactly as it did.

## Off, inside or outside — not a switch

Which position reads better is not something the code can decide for somebody. The notch is small, the two arcs are close together at any size, and whether the inner gap or the bezel margin is the clearer place depends on the size the notch is set to and on the eyes looking at it. Measured, from the ring's centre:

| | arc | clears |
|---|---|---|
| inside | 9.59 – 11.47pt | 0.94pt from the glyph, 1.03pt from the working indicator |
| outside | 23.50 – 25.38pt | 1.50pt from the track, 9.59pt from the bezel |

**No existing `NotchLayout` constant is touched** — `git diff` on that file deletes zero lines. The new radii are placed against what is already in the circle rather than quoted from the design frame, which draws one ring, so every measurement there still answers to `docs/design/frame-124-hover-tooltip.png`.

## Two things found by looking rather than by testing

**A week at 0% drew nothing.** Codex opened its weekly at 0%, the arc had zero length, and with nothing behind it that is indistinguishable from the feature being broken — which is exactly how it was reported to me. The second ring now has a track of its own, for the same reason the headline has one.

**Inside, it yields to the working indicator.** The two want the same band, 1.03pt apart, one of them spinning. Rather than shave both until neither is legible, the transient fact wins while a provider is working and the week stays a hover away. Outside there is no contest and nothing is given up.

## Which window

Declared by the provider, like `headlineID` and for the same reason: everyone who has a weekly window calls it something different — `weekly_all`, `secondary`, `weekly`, `gemini-weekly` — and a rule that guessed from `duration` would silently skip whichever provider had not filled that field in, with nothing on screen to say why.

Never the window the headline is already drawing. Antigravity picks its headline by whichever limit is tightest and so lands on its own weekly some of the time; two rings reporting the same number reads as a second fact that happens to agree rather than as one fact drawn twice.

Populated for Claude, Codex, GLM, OpenCode, Ollama, Command Code and Antigravity. Not for Grok, whose `credits` window *is* the weekly one and is already the headline.

## Test plan

- [x] `make test` passes with no signing override.
- [x] Clearances are pinned as clearances rather than as numbers: inside clears the glyph and the working indicator, outside clears the track and the bezel, and the weekly arc is thinner than the headline's.
- [x] Rendered, not calculated, where the question is whether something appears — and with two different measures, because each is blind to one case. The arcs are drawn over the notch's own opaque black, so counting ink cannot see them: it answers the same number to seventeen decimal places whether the ring is there or not. A 0% arc carries no hue, so counting colour cannot see its track. Both of those were caught by watching a test pass that should have failed.
- [x] Both regressions fail their tests when reverted — removing the track makes the empty-week test report an identical pixel count, which is the failure it exists to catch.
- [x] Live on this machine, through the running app: Claude `session` → `weekly_all` at 47%, Antigravity `gemini-5h` → `gemini-weekly` at 46% (agreeing with `agy`'s own panel, which reports 53.57% remaining), Codex `primary` → `secondary` at 0% and visible by its track.
- [ ] **Only three providers are connected here**, so GLM, OpenCode, Ollama and Command Code are wired but never drawn by me. Their `weeklyID` is a one-line declaration each and resolves by id, so the risk is naming rather than logic — but I have not seen them.
- [ ] Not tried with a hardware notch or on a second display.

## One hole I would rather name than leave for you to find

When a provider's headline lands on its own weekly window, the second ring correctly disappears — but Antigravity has four windows, so there is always another pair worth showing, and it shows nothing instead. Making the second ring "the principal window that is not the headline" would close that, at the cost of the name: it stops being a weekly ring. I left it as it is rather than widen the idea without asking; say the word if you would rather have it.

### DIFF
--- a/Sources/App/AppDelegate.swift
+++ b/Sources/App/AppDelegate.swift
@@ -293,6 +293,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                 .sink { [weak fleet] in fleet?.apply(accentColor: $0) }
                 .store(in: &cancellables)
 
+            preferences.$weeklyRing
+                .receive(on: RunLoop.main)
+                .sink { [weak fleet] in fleet?.apply(weeklyRing: $0) }
+                .store(in: &cancellables)
+
             preferences.$disconnectedProviders
                 .receive(on: RunLoop.main)
                 .sink { [weak store] in store?.disconnected = $0 }
@@ -463,6 +468,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         fleet.apply(scale: preferences.notchScale)
         fleet.apply(resetTimeFormat: preferences.resetTimeFormat)
         fleet.apply(accentColor: preferences.accentColor)
+        fleet.apply(weeklyRing: preferences.weeklyRing)
         fleet.show()
     }
 
--- a/Sources/Features/ProviderRing.swift
+++ b/Sources/Features/ProviderRing.swift
@@ -21,6 +21,11 @@ struct ProviderRing: View {
     /// A fetch this cell asked for, in flight.
     var isRefreshing: Bool = false
     var localPerformance: LocalModelPerformance?
+    /// The weekly limit, when the provider has one. Nil is the ordinary case
+    /// for a provider with a single window, and draws nothing.
+    var weeklyFraction: Double?
+    /// Where the user asked for it, if at all.
+    var weeklyRing: WeeklyRing = .off
 
     @Environment(\.accessibilityReduceMotion) private var reduceMotion
     @Environment(\.codenotchReduceTransparency) private var reduceTransparency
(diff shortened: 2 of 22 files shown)


mulhamna:
The red check here is `EdgeArrivalTests.testItLandsFoldedAndThenOpens`, and I do not think it belongs to this branch.

It fails the same way on `feature/success-state`, in three separate runs — 34357168226, 34353128652 and 34342954614 — which is a branch that has nothing to do with this one. On this run it is also the *only* failure: every one of the six tests this PR adds passed on CI, including the two that render the view and count pixels, which were the ones I expected to be environment-sensitive.

There is no mechanism I can find either. The test never touches the weekly ring, and the setting defaults to `off`, in which state the view adds nothing at all.

For what it is worth on the cause: the test waits for a *transient* state — `panelAlphaForTesting` dropping below 1 and then returning to it — by polling every 0.02s for up to 3s. If the crossfade completes between two polls, "it never went away" is reported for a transition that did happen. A slower or headless runner widens that window rather than causing it.

I could not confirm by re-running: `gh run rerun` answers `Must have admin rights to Repository`. A re-run from your side would settle it either way, and I would rather you decided that than have me push an empty commit to force one.

Happy to send a separate PR making that test wait for the transition to have completed rather than trying to catch it mid-flight — but it is your test and you may have had a reason for watching the intermediate state, so I have not touched it.

PR #112, opened 2026-09-10
Windows: read Antigravity quotas through the official CLI

Sargares22:
## Problem and change

Windows users with the official Antigravity CLI installed can now read quota while the IDE is closed. The adapter prefers the standalone `agy --print /usage` report and runs it through native ConPTY, without a personal PowerShell wrapper or an additional service.

Refresh runs at startup and on hover/manual requests when readings are at least five minutes old. Failed attempts are throttled too. The runner bounds its process tree and output, and cache writes replace a synced temporary file. Errors retain the last reading as stale instead of falling back to internal APIs.

Installations without the CLI retain the existing provider behavior, including their periodic refresh cadence: hover only wakes the CLI worker. Source selection occurs at startup, so installing or removing the CLI requires restarting Codenotch. The text report format remains a compatibility boundary; this change does not claim a blanket ToS assurance.

No UI layout changes are included.

## Validation

- Windows 11 ARM64: release build passed.
- `cargo test --release --locked`: 18 offline tests, live integration test ignored by default.
- Explicit live integration run: all 19 tests passed, including the signed-in official `/usage`.
- Covered malformed quota output, missing executable, Windows quoting, bounded output, timeout and failed cache replacement.
- Native app displayed four live quota windows without clipping. HTML fixture checks covered hover/leave and bounds at DPR 1 and 2.
- Windows x64 and macOS were not tested. The CONTRIBUTING.md `make test` requirement needs macOS/Xcode and was not run on this Windows host. GitHub currently reports no check results; its workflows cover macOS, not Windows. The existing unused-function warning remains.

The `STARTF_USESTDHANDLES` setting is deliberate: on the tested host, omitting it sent child output to the parent instead of ConPTY. The runner regression test covers capture of both short and larger output.


RawJat:
Ran this on Windows 11 x64, since x64 is listed as untested. cargo test: 18 passed, 1 ignored. The ignored one is live_quota, which needs a signed-in agy, and I don't have one. Both test_bounded_runner_and_no_deadlock and test_timeout_cleanup_terminates_process pass, so the ConPTY capture with STARTF_USESTDHANDLES holds on a second architecture. cargo clippy --all-targets adds no new warnings; the one in antigravity.rs is the old chunks_exact, now at line 312.

For anyone following: this covers the CLI. IDE-only installs take a different path, #130, fixed in #132. The two don't overlap, but both touch present(). If #132 lands first, legacy_present()'s state_root() call becomes !state_roots().is_empty(). Happy to rebase #132 instead if this goes first.

PR #116, opened 2026-09-10
Fix the ad-hoc signing fallback never triggering

and-human24:
`grep -c` prints a count, so the shell output is `0` when no Developer ID
certificate is present — never the empty string the `ifeq` compares against.
The guard is therefore false on exactly the machines it exists for, `DEV_SIGN`
stays unset, and Debug falls through to project.yml's release identity.

On any machine without your certificate, `make build` / `make test` / `make run`
all fail:

```
No signing certificate "Developer ID Application" found: No "Developer ID
Application" signing certificate matching team ID "6WFPL8B9FB" with a private
key was found. (in target 'Codenotch')
```

That is the opposite of what CONTRIBUTING.md promises. Dropping `-c` restores
it: absent certificate means empty output, the guard matches, Debug ad-hoc
signs itself.

**Tests:** no new test — this is `make` conditional logic, out of reach of the
Swift suite. Verified by building clean `main` on a machine without the
certificate (fails as above) and this branch (`make test`, 552 pass).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,12 @@ DEST    := platform=macOS,arch=arm64
 # changes nothing: project.yml's stable identity is what keeps a keychain
 # "Always Allow" grant alive across rebuilds, and forcing ad-hoc there would
 # throw that away and bring the prompt back on every `make run`.
-ifeq (,$(shell security find-identity -v -p codesigning 2>/dev/null | grep -c "Developer ID Application"))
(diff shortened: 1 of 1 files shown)


vinzdg:
Thank you — this one is already fixed on `main`: the Makefile moved from `grep -c` to `grep` for exactly the reason you give (a count prints "0", never the empty string the guard compares against), and it has since been restructured around `HAS_DEVELOPER_ID`. Closing as already done rather than merging over it.

PR #117, opened 2026-09-10
Stop the keychain password prompt from coming back

and-human24:
A keychain item has two gates. The access list is what "Always Allow" writes
to. The **partition list** is written by no dialogue, and an app outside it is
refused before the access list is consulted — so approving the prompt records a
grant that is discarded on the next read.

Claude Code recreates its keychain items on every token rotation rather than
updating them (`cdat == mdat` on each new item), and a fresh item's partition
list holds only `apple-tool:`. That evicts any app with a Team ID — release
build included — roughly every eight hours:

```
securityd: ACL partition mismatch: client teamid:<TEAMID> ACL ( "apple-tool:" )
securityd: displaying keychain prompt for /Applications/Codenotch.app
```

Filed upstream as anthropics/claude-code#22144 (by the author of another menu
bar app) and closed as not planned, so it has to be handled here.

**Fix:** never raise the dialogue from a poll, and recover without it.
`SecKeychainSetUserInteractionAllowed(false)` is what legacy login-keychain
items honour — `kSecUseAuthenticationUI` alone governs the data-protection
keychain and is silently ignored by anything carrying a partition list, which
is every item involved. On a refusal, read through `/usr/bin/security`, which
is Apple-signed and therefore *is* `apple-tool:`:

```
codenotch:usage  Claude Code-credentials-<hash> read via the security tool after a partition refusal
```

A refusal even that cannot rescue stays `.accessDenied`, with copy naming the
one thing that restores direct access instead of a button that cannot.

Also corrects the CONTRIBUTING.md paragraph describing this as a
development-only annoyance.

**Tests:** 552 pass; the status copy has one. The fallback spawns a process
against the real keychain and the suppression is a process-wide macOS setting,
so neither is reachable in-process — `WebSessionProvider` has no unit tests for
the same reason. Verified on a machine that had the prompt once a minute: zero
prompts over seven minutes of active polling, including the partition refusal
above recovered silently.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,12 +10,25 @@ make run                 # build and launch
 ```
 
 None of these need an Apple Developer account. `xcodebuild` ad-hoc signs a
-Debug build automatically, which is enough to run and debug locally. The one
-thing an unsigned build can't do is keep a keychain "Always Allow" grant across
-rebuilds — Claude Code's and Antigravity's credentials are guarded by an ACL
-keyed on the signing identity, and an ad-hoc identity changes every build. In
-practice this means the keychain prompt reappears each time you rebuild during
-development; that's expected and doesn't affect anything else.
+Debug build automatically, which is enough to run and debug locally. An ad-hoc
+identity changes every build, so a keychain "Always Allow" grant does not
+survive a rebuild and the prompt reappears during development.
+
+That prompt is no longer shown, in Debug or in a release build. It was never
+only a development annoyance: a keychain item has an access list, which is what
+"Always Allow" writes to, and a *partition list*, which nothing in the GUI ever
+writes to. An app outside the partition list is refused before the access list
+is consulted, so approving the dialogue is good for exactly one read. Claude
+Code recreates its keychain items on every token rotation rather than updating
+them, and a freshly created item's partition list holds only `apple-tool:` —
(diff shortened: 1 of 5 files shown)


vinzdg:
Thank you — this landed in `main` as 77859e9, ported from 1.5.0 with you credited as co-author. Your diagnosis was right and the core is yours: background reads switch interaction off and fall back to `/usr/bin/security` after a refusal. `fix-keychain-partitions.sh` is in as written.

Three changes, each because of something that arrived after 1.5.0:

- **"Allow access…" still prompts.** 1.8 has a Settings button whose whole job is to ask macOS again. Disabling interaction for every read would have left it unable to show anything. That button now grants the *next* read permission to prompt — spent by that read whatever the answer, so a Deny can't leave it for a poll, and it lapses after a minute. The automatic `forgetCached` paths (a rejected token, the token refresher) grant nothing.
- **The switch is locked and restored, not forced back on.** Two profiles reading at once could interleave the save and restore and leave interaction off for good.
- **The fallback reads the item the direct read chose.** Multiple profiles means multiple `Claude Code-credentials-*` items, so `KeychainItem.Match` now carries its service name.

The status message points at Settings' Allow access rather than the script, since the script isn't in the installed app — it's referenced in CONTRIBUTING instead.

PR #118, opened 2026-09-10
Stop the back-off beating against the refresh tick

and-human24:
The usage endpoint answers a 429 with a 60s hint, and `UsageStore` also ticks
every 60s. Running at the same period, a tick lands a few milliseconds *before*
the window opens:

```
codenotch:usage  claude-<profile> failed: rateLimited(retryAfter: 0.015)
codenotch:usage  claude-<profile> failed: rateLimited(retryAfter: 0.42)
```

Refusing there saves 15ms and costs a whole refresh interval, because the
caller is a timer rather than a retry loop — the next attempt is 60s away, not
a moment away. A 60s penalty becomes 120s, every other tick produces nothing,
and a rate-limited provider takes twice as long to recover as the server asked.

Treat a window with under a second left as open.

**Tests:** 555 pass. Extracted as a pure `shouldHoldOff(until:slack:now:)` so
the resonance is testable without a timer or a live endpoint; three cases added
to `ClaudeOAuthProviderTests`.

### DIFF
--- a/Sources/Providers/ClaudeOAuthProvider.swift
+++ b/Sources/Providers/ClaudeOAuthProvider.swift
@@ -56,8 +56,28 @@ actor ClaudeOAuthProvider: UsageProvider {
         self.retryNoEarlierThan = archive.loadBackoffUntil(providerID: profile.id)
     }
 
+    /// How close to expiry a back-off counts as already expired.
+    ///
+    /// The server hands back a 60s hint and `UsageStore` also ticks every 60s,
+    /// so the two run at the same period and the tick lands a few milliseconds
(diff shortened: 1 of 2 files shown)


vinzdg:
Thank you — this landed in `main` as c4002ce, ported rather than merged because the branch was written against 1.5.0 and no longer applied. The fix is yours as written, with you credited as co-author.

One adaptation: since 1.8 the provider tries the `claude` CLI before the back-off check (the CLI doesn't share the endpoint's rate limit), so the slack now applies where that check lives, after the CLI path. Your three tests are in unchanged. I left out the empty `Tests/ClaudeOAuthTests.swift` the branch also added.

PR #119, opened 2026-09-10
Retry a keychain read that failed for reasons outside the item

and-human24:
`CredentialCache` remembers a failed read against the item's modification date.
That is right for a failure the item caused — a payload that will not decode
gives the same answer every time, so re-reading an unchanged item is noise.

It is exactly wrong for a failure the item had nothing to do with. macOS
declining to show a prompt does not move `mdat`, so the error stays pinned to a
version of the item nothing will ever write again and the read is never
retried. One blip five seconds before a clamshell sleep produced 2h42m of
"signed out" over a token that was valid throughout — one logged error, then
fifty silent replays, ending only when the app was quit.

Environmental failures now fall through to `retryAfterFailure`, which does
expire. `-60008` (`errAuthorizationInternal`) joins `-25320` as transient for
the same reason: both mean "not right now", not "signed out" — and reporting a
sign-out also makes `supersedesHistory` discard the archived reading.

**Tests:** 555 pass. Three cases added to the existing `CredentialCacheTests`,
including the pinned-error regression and the case the stamp is still right for.

### DIFF
--- a/Sources/Providers/ClaudeCredentials.swift
+++ b/Sources/Providers/ClaudeCredentials.swift
@@ -126,7 +126,12 @@ struct ClaudeCredentials {
     /// would be needed. Security doesn't export a named constant for it, so
     /// the raw value is what there is to check.
     static func wasTransient(_ status: OSStatus) -> Bool {
-        status == -25320   // errSecInDarkWake
+        status == -25320       // errSecInDarkWake
+            // errAuthorizationInternal. What a refusal looks like when macOS
+            // decided a prompt was needed and there was no way to show one —
+            // observed five seconds before a clamshell sleep. The credential is
+            // untouched, so this is "not right now", not "signed out".
+            || status == -60008
     }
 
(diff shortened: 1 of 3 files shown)


vinzdg:
Thank you — the `-60008` part landed in `main` as b730437, ported from 1.5.0, with you credited as co-author. An unshowable prompt now ages the reading like a dark wake instead of reading as signed out.

The rest wasn't taken, and I want to be clear why so it doesn't look like it was missed:

- **Retrying environmental failures on a timer** — `main` already does this, the other way round: `CredentialCache` retries every failure after `retryAfterFailure` unless it's classed permanent (`isPermanentFailure`), and `testATransientFailureIsRetriedAfterTheWindowEvenWithTheSameMdat` covers the case you describe.
- **Reclassifying `accessDenied` as retryable** — deliberately not taken. `main` keeps an explicit Deny permanent, because retrying a refusal on a timer is the prompt loop that cache exists to stop.

PR #120, opened 2026-09-10
Keep the last reading when the owning app empties its own credential

and-human24:
> **Stacked on #117** — it needs the shared decode path extracted there, so this
> branch contains that commit too. Merge #117 first and this reduces to a single
> commit touching 4 files.

Claude Code rewrites its keychain item with an empty access token, no refresh
token and `expiresAt: 0`. It decodes perfectly and is worth nothing. Reading it
reported `needsAuth`, which means "nobody ever signed in here" — and
`supersedesHistory` treats that as reason to erase the archived reading.

So a glitch in another program empties the notch. Three profiles went at once,
within two seconds, after Claude Code auto-updated overnight and the machine
woke:

```
security[94183] integrity: no previous integrity acl exists; making a new one   (x6)
codenotch:usage  Claude Code-credentials-<hash> holds an emptied credential
```

That is anthropics/claude-code#19456, closed as not planned; nothing here can
prevent it.

What this side can do is tell the two states apart. An item that exists and has
been hollowed out is evidence somebody *was* signed in and did work, so the
numbers from before are still true about the account. `signedOutByOwner` keeps
them, shows them as stale, and says what happened — while `needsAuth` still
blanks a profile nobody has ever used.

**Tests:** 556 pass. Four cases added beside the existing refusal tests:
history retention, that `needsAuth` still clears, the error/status mapping, and
the copy.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,12 +10,25 @@ make run                 # build and launch
 ```
 
 None of these need an Apple Developer account. `xcodebuild` ad-hoc signs a
-Debug build automatically, which is enough to run and debug locally. The one
-thing an unsigned build can't do is keep a keychain "Always Allow" grant across
-rebuilds — Claude Code's and Antigravity's credentials are guarded by an ACL
-keyed on the signing identity, and an ad-hoc identity changes every build. In
-practice this means the keychain prompt reappears each time you rebuild during
-development; that's expected and doesn't affect anything else.
+Debug build automatically, which is enough to run and debug locally. An ad-hoc
+identity changes every build, so a keychain "Always Allow" grant does not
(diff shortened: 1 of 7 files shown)


vinzdg:
Thank you — this landed in `main` as ceb3b2d, ported from 1.5.0, with you credited as co-author. `signedOutByOwner` is in as you designed it and your four tests are unchanged.

Two adaptations: it no longer depends on #117 — `main` decodes inline, so the empty-token check sits right after the payload decodes. And the message is a single localizable string rather than rewriting the English "Sign in to" into "Sign in again to", since 1.8 ships French, Portuguese and Chinese and that substring replacement would have silently done nothing in them.

PR #143, opened 2026-09-11
Store the providers that are on, not the ones that are off

DrSmoothl:
## Summary

Codenotch used to remember which providers were switched off. Anything not on that list was treated as on, so a new install — and every newly added provider — lit up rings the user had never asked for. If you only subscribe to one or two tools, you had to go through Settings and turn the rest off.

This stores the providers that are on instead.

A first install starts with Claude and Codex, which is probably what most people actually have. Everything else stays off until you switch it on. An existing install keeps the rings you already chose. A provider added in a later version stays off; a newly discovered Claude or Codex account still defaults on.

## Test plan

- [x] First install: only Claude / Codex connected
- [x] Upgrade: previously-on providers stay on
- [x] Toggle Cursor on, relaunch, it stays on
- [x] A newly discovered Claude profile defaults on without turning Cursor on

### DIFF
--- a/Sources/App/AppDelegate.swift
+++ b/Sources/App/AppDelegate.swift
@@ -118,24 +118,26 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             Log.usage.info("codex profiles: \(self.codexProfiles.map(\.displayPath).joined(separator: ", "), privacy: .public)")
             let claudeProviders = claudeProfiles.map { ClaudeOAuthProvider(profile: $0) }
             self.claudeProviders = claudeProviders
+            let allProviders: [UsageProvider] = claudeProviders
+                + [CursorLocalProvider()]
+                + codexProfiles.map { CodexLocalProvider(profile: $0) }
+                + [AntigravityProvider(),
+                   GLMProvider(), GrokLocalProvider(), DevinLocalProvider(), OpenCodeProvider(),
+                   CommandCodeProvider(), GitHubCopilotProvider(),
+                   OllamaLocalProvider(endpoint: URL(string: preferences.ollamaEndpoint)!),
+                   LMStudioLocalProvider(endpoint: URL(string: preferences.lmstudioEndpoint)!),
+                   OllamaProvider(),
+                   // A closure, not the value: the provider is an actor and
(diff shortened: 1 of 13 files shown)


vinzdg:
The inversion itself is right and the migration reads well, but it hides every local model cell on upgrade, so I've held it back.

`UsageStore.knownIDs` is provider ids **plus** `localModelSummaries` — each loaded Ollama or LM Studio model is its own id, like `ollama-local:model:qwen3:8b`. `AppDelegate` then sets `store.disconnected = preferences.disconnectedIDs(among: store.knownIDs)`, which is everything `isConnected` says no to. But `reconcile` is only ever handed provider ids (`allProviders.map(\.id)`), so no model cell is ever in `connectedProviders` — and once the `connected` key exists, absence means off.

Under the old off-list a model cell was visible unless explicitly disabled. After this, every one is hidden the moment someone updates, and each newly loaded model stays hidden until switched on by hand.

Reproduced against your branch merged onto `main`:

```swift
d.set([String](), forKey: "disconnectedProviders")   // a 1.9 install, nothing hidden
let p = Preferences(defaults: d)
p.reconcile(discoveredIDs: ["claude", "codex", "ollama-local"])   // as AppDelegate calls it
p.disconnectedIDs(among: ["claude", "codex", "ollama-local", "ollama-local:model:qwen3:8b"])
// contains "ollama-local:model:qwen3:8b"
```

The awkward part is that "absent means on" and "absent means off" cannot both hold in one list, so model cells probably need to keep an explicit off-list of their own — which is what the `disabledModels` comment your diff removes was doing. That is your design call, so I have not guessed at it.

Everything else — the rename, `pendingHidden`, the one-time inversion, the seen-set — looks sound, and the tests you added cover the provider paths well.

DrSmoothl:
You're right — one list cannot mean both.

Providers stay stored as the ones that are on. Loaded-model cells keep an explicit off-list of their own, which is what the old disabledModels comment was doing. A model Ollama or LM Studio loads later stays visible unless someone hid it.

On upgrade, model ids that were already on hiddenProviders stay hidden; everyone else stays shown. Your repro no longer includes ollama-local:model:qwen3:8b, and switching a provider on or off no longer hides models that nobody hid.

PR #149, opened 2026-09-11
Add Kimi provider, session tracking, and exact-tab session focus

8nevil8:
## Summary

Adds **Kimi** (Kimi Code CLI) as a provider — usage rings for the 5-hour rate window and weekly quota, plus live session tracking — and makes **every session row in the tooltip clickable**, jumping to the *exact terminal tab* the session runs in, not just the app.

## Changes

**Add Kimi usage rings** (`Sources/Providers/Kimi*.swift`)
- Reads the CLI's own OAuth session from `~/.kimi-code/credentials/kimi-code.json` (`KIMI_CODE_HOME` honoured) and queries the same `/usages` endpoint the CLI's `/usage` leads with.
- Headline ring = rolling 5-hour rate window; second ring = weekly quota; membership level shown as the plan. Read-only, same bargain as Grok: the 15-minute token is the CLI's to refresh, so an expired one reports `credentialExpired` rather than racing the CLI for the file.
- Defined geometric-K glyph (same approach as GLM's Z), README row, setup copy.

**Follow running Kimi Code sessions** (`Sources/Sessions/KimiActivityMonitor.swift`)
- Kimi writes no live-session registry, so discovery finds `kimi-code` processes and matches each one's cwd to `session_index.jsonl`; state comes from the session's `wire.jsonl` (turn boundaries and `approval.requested`/`resolved` with ms timestamps): busy / waiting / success / idle, with loop records counting as busy so a long turn's own `turn.prompt` scrolling out of the tail doesn't misreport.
- The pid goes into the session, so rows can be jumped to; the second line names the owning app, the way Claude's rows name the surface.

**Jump to a session's exact terminal tab on click** (`TerminalTabFocus.swift`, `SessionFocus.swift`, tooltip/notch wiring)
- Session rows are clickable for any provider whose sessions carry a pid (Kimi, Claude, Grok).
- **cmux**: selects the exact surface found by the `CMUX_SURFACE_ID` the session's own process tree carries in its environment — its socket CLI refuses clients that aren't inside a cmux terminal session (manaflow-ai/cmux#3089), so the AppleScript interface answers instead, with working-directory matching as fallback. **Terminal.app / iTerm2**: tab matched by tty. Everything else keeps the plain app raise.
- The completion-peek click uses the same jump. Claude sessions whose state comes from the transcript now keep their pid, so those rows are clickable at all. Card clicks no longer fall through to the pin toggle.

## Impact

Before: hovering showed sessions but clicking did nothing; the peek click raised only the app. After: one click lands you in the exact tab (verified on cmux with multiple same-directory Claude sessions, which cwd matching alone could not tell apart). AppleScript terminals ask for a one-time Automation consent; cmux needs none beyond that.

## Verification

- New tests: `Tests/KimiUsageTests.swift`, `Tests/KimiActivityTests.swift`, `Tests/TerminalTabFocusTests.swift` (pinned live-recorded payloads; wire state machine; env/tty/path parsing).
- Verified end-to-end on a live machine: real fetch against a signed-in Kimi account, session state tracking, and repeated exact-tab jumps (including same-cwd disambiguation via `CMUX_SURFACE_ID`).
- Note: this machine has no full Xcode, so `make test` could not be run here; the suites above were mirrored in a compiled harness (43/43 pass) and the app module typechecks clean. CI should confirm.

## Checklist
- [x] Self-reviewed
- [x] Manual testing performed
- [x] Documentation updated (README provider table)
- [x] No breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ A Windows port — Rust/Tauri 2, same design and providers — lives in [`window
 | **OpenCode** | official | The Go plan's official usage endpoint, with the `opencode-go` key OpenCode itself stores on sign-in. |
 | **Command Code** | official | The GOAT plan's `/alpha` billing endpoints, with the key the Command Code app writes to `~/.commandcode/auth.json`. |
 | **GitHub Copilot** | official | GitHub's Copilot quota endpoint, authenticated with the GitHub CLI session already on the Mac (`gh auth login`). |
+| **Kimi** | official | The Kimi Code CLI session in `~/.kimi-code/credentials/kimi-code.json`, against the same `/usages` endpoint the CLI's `/usage` asks. Shows the 5-hour rate window and the weekly quota. |
 
 Most providers borrow a credential or session from a tool already on your Mac.
 DeepSeek is the explicit browser-login exception: it never reads a browser's
--- a/Sources/App/AppDelegate.swift
+++ b/Sources/App/AppDelegate.swift
@@ -124,7 +124,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
                     + codexProfiles.map { CodexLocalProvider(profile: $0) }
                     + [AntigravityProvider(),
                        GLMProvider(), GrokLocalProvider(), DevinLocalProvider(), OpenCodeProvider(),
-                       CommandCodeProvider(), GitHubCopilotProvider(),
+                       CommandCodeProvider(), GitHubCopilotProvider(), KimiProvider(),
                        OllamaLocalProvider(endpoint: URL(string: preferences.ollamaEndpoint)!),
                        LMStudioLocalProvider(endpoint: URL(string: preferences.lmstudioEndpoint)!),
                        OllamaProvider(),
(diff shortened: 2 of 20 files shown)


vinzdg:
Nice work — one thing to fix before this goes in.

`TerminalTabFocus` builds AppleScript by string interpolation. The working directory is correctly escaped through `appleScriptEscaped`, but two other values aren't:

```swift
selectCmuxTerminal(matching: "id of term is \"\(surface)\"")   // line 28
…
if tty of t is "/dev/\(tty)" then                              // lines 41, 57
```

`surface` comes from `cmuxSurfaceID(of:)`, which reads the process's environment, and `tty` from the process table. An environment variable is set by whoever launched the process, so a value containing a quote would close the string literal and the rest would run as AppleScript, as this user. Same class you already guard against for `cwd` — `appleScriptEscaped` on both should be all it needs.

The rest read fine to me: the Kimi credential is read from its own config, and the only endpoint is `api.kimi.com`.

PR #150, opened 2026-09-11
Add a Models pane for choosing what the notch shows

giusfds:
## Why

The list of assistants Codenotch can read keeps growing, and today the only way
to take one out of the notch is to disconnect it in **Accounts** — which stops
its polling and forgets its last good reading. "I don't want to look at this
one" and "I don't want Codenotch reading this account" are different wishes,
and only the second one is currently expressible.

## What this adds

**Settings → Models**: one switch per assistant, and one switch per loaded
Ollama / LM Studio model grouped under its runtime. Each runtime also has a
switch for all of its cells at once; turning it off keeps the individual model
choices for when it comes back on.

Visibility is a display projection and nothing else. A hidden item keeps its
credential, its polling, its archived readings and its position in the order —
`UsageStore` now keeps the arranged cell list (`notchCells`) and filters it into
`notchSnapshots`, so showing something again restores its place without a fresh
poll. Hiding a model does not unload it from Ollama or LM Studio.

## Built to survive the next provider

Nothing in the pane is keyed to a provider name. It groups whatever
`providerSummaries` reports by `ProviderKind`, so an assistant added in a later
release appears in Models on its own, with no new case to write. The single
place that has to recognise a local model cell matches the shared
`<runtime>:model:<model>` id shape that `UsageModel` and `LMStudioMetrics`
already build, rather than a list of known runtimes — adding the next local
runtime touches none of this. There's a regression test using a runtime that
does not exist (`mlx:model:phi-4`) to keep it that way.

## Compatibility

The per-model switch in **Accounts** stays where it is and now writes the new
preference, with a caption pointing at Models when a row is hidden. Local model
visibility used to be stored alongside the connection choices in
`hiddenProviders`; showing a model from either pane clears that legacy entry
without reconnecting its runtime, so nobody's existing choices are lost or
silently turned into sign-outs.

Reordering stays in **Accounts** and is unchanged: drag a visible row by its
handle, and a provider reconnected there still joins the end of the list.

## Tests

`Tests/ModelVisibilityTests.swift`:

- visibility persists separately from connections and survives a relaunch;
- hiding never touches readings, the archive, the order, or an in-flight fetch;
- a hidden provider still loads its archive and still refreshes on schedule;
- runtimes and their models hide independently while model discovery continues;
- the legacy migration restores a model belonging to an unknown runtime.

## Copy

All new strings go through `L10n.t` and carry fr / ja / pt-BR / zh-Hans
translations in `Sources/Localizable.xcstrings`. README updated for the new pane.

## What I could not verify

I don't have Xcode on this machine (Command Line Tools only), so `make test`
cannot build the app target locally. The changed files pass `swiftc -parse` and
`xcodegen generate` picks up both new files, but CI on this PR is the first real
build. Happy to fix anything it turns up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,12 +71,20 @@ Most providers borrow a credential or session from a tool already on your Mac.
 DeepSeek is the explicit browser-login exception: it never reads a browser's
 cookies or credentials, and only makes requests after you choose **Sign in to
 DeepSeek** from Codenotch.
-Ollama Cloud accepts an API key in Settings. Switching a provider off stops its
-usage polling and forgets its readings; borrowed accounts stay signed in to
-the tools that own them.
+Ollama Cloud accepts an API key in Settings. Disconnecting a provider in
+**Settings → Accounts** stops its usage polling and forgets its readings;
+borrowed accounts stay signed in to the tools that own them.
+
+**Settings → Models** lets you show or hide assistants and loaded Ollama or
+LM Studio model cells to keep the notch focused. These choices are remembered
+across launches. Hidden items keep their connections, monitoring and readings.
+Each local runtime also has a switch for all of its model cells; switching it
+off preserves the individual model choices for when you show it again. Hiding
+a model does not unload it from Ollama or LM Studio.
 
 **Local Ollama is detected automatically.** Configure its address or stop monitoring in **Settings → Ollama**.
-Each loaded model gets a notch cell; reorder or hide it in **Settings → Accounts**.
+Each loaded model gets a notch cell; hide or show it in **Settings → Models**
+and reorder visible cells in **Settings → Accounts**.
 Hover for RAM/VRAM, unload time, context limit and quantization.
 
 For generation speed (**tok/s**) and live **Thinking**, enable **Measure speed and thinking**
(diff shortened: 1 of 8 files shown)


vinzdg:
Thanks for this — holding it back for now because it isn't green on its own branch, and I'd rather not land a red `main`.

Four assertions fail in your own `ModelVisibilityTests`, reproduced on `refs/pull/150/head` in a clean worktree with nothing else applied:

```
testHiddenProviderLoadsItsArchiveAndContinuesToRefresh
  :91  XCTAssertTrue failed
  :92  XCTAssertNotNil failed
  :96  XCTAssertEqual failed: ("nil") is not equal to ("Optional(…)")
testHidingAnInFlightProviderKeepsAndArchivesItsResult
  :118 XCTAssertNotNil failed
```

They look like the same thing twice: a hidden provider's archived reading comes back `nil`, so the archive either isn't being written for hidden providers or isn't being read back for them.

Worth knowing when you rebase: #143 is open and changes the same area from the other side — it stores the providers that are **on** rather than the ones that are off. Whichever lands first, the other will want a look.

vinzdg:
Checking back in: the branch head is still `ac3eb85`, the same commit I ran two days ago, so the four `ModelVisibilityTests` failures I posted still stand and I'm leaving this unmerged.

One thing has changed underneath it that you'll want to know before you pick it up again: #143 is now merged. It stores enabled providers as an on-list, and keeps loaded-model cells on an explicit off-list of their own (`Preferences.isConnected` / `disconnectedIDs(among:)`, with `UsageStore.disconnected` as the computed set). A Models pane for choosing what the notch shows sits directly on that, so a rebase onto `main` is probably the natural moment to wire it to #143's API rather than the old one — and the failing archive assertions may look different afterwards.

vinzdg:
Closing this for now, since it's been waiting on the issues mentioned above and main has moved on a lot since. Thanks for the work, @giusfds. If you'd like to pick it back up, feel free to rebase on current main and open a fresh PR.

PR #152, opened 2026-09-11
Show on each Windows ring the window the Mac app shows

RawJat:
The ring on the Windows notch shows whichever window is most used, and so does the tray icon since #115. So Claude's ring switches to the weekly limit as soon as it's fuller than the session, and disagrees with `/usage` and with the Mac app, which always lead with the current session. The Mac app dropped that rule on purpose (TASKS.md, "The headline that moved when a window reset").

Each ring now shows the window the macOS provider declares as `headlineID`:

| provider | ring shows |
|---|---|
| Claude | `session` |
| Codex | its first window (`primary`) |
| Cursor | `included`, else `api` |
| Antigravity | the Mac app's **Notch reads** (Automatic / 5-Hour / Weekly), within **Model data** (Gemini / Claude and GPT). Same defaults: Automatic and Gemini |

If the declared window is missing from a reply, the ring shows `—` rather than putting another window in its place.

Settings follows the Mac app:
- the window list under each ring is gone
- Antigravity's row gets the two pickers
- a tray slot shows its provider's ring
- a `window` saved by the previous build is ignored

The rule lives in `ring_window` in `main.rs`, which the tray uses. `headlineOf` in `notch.html` mirrors it.

Rebased on `main` after #141 landed Russian. The window lists this PR removes took their dictionary entries with them, and the two pickers it adds go through #141's `ui()` helper. Their Russian comes from the Mac's own catalog — "Notch reads" is «Показания панели», and the lane and model names are already translated there. "Model data" has no entry in that catalog, so it stays English here exactly as it does on the Mac.

Two captions shrink with the feature they described, and keep their Russian: #141 matches static text exactly, so a reworded caption silently loses its translation. Both RU_STATIC keys move to the new wording — one of them simply drops the sentence about the window list this PR removes.

## Test plan

- [x] `cargo test`: 34 pass, 8 of them new:
  - Claude's session beats a fuller week
  - a missing session shows a dash
  - Codex's and Cursor's windows
  - Antigravity's Gemini-only default
  - Notch reads
  - the `agy` CLI's lane names
  - a spent lane
  - a reading that is only a request count
- [x] `cargo clippy --all-targets`: the same 30 warnings as `main`.
- [x] Both pages' scripts parse (`node --check`).
- [x] Settings rendered in Russian in headless Edge, bridge stubbed: the pickers read «Показания панели» → «Автоматически» and «Модели Gemini», and both reworded captions are translated.
- [x] Live on Windows 11:
  - Claude's ring matches "Current session".
  - Settings → Notch shows the pickers only under Antigravity, and changing them moves its ring and the tray icon.
  - The tray picker lists providers with their ring numbers.

### DIFF
--- a/windows/codenotch/src/config.rs
+++ b/windows/codenotch/src/config.rs
@@ -6,15 +6,11 @@ use std::path::PathBuf;
 pub const SCALE_MIN: f64 = 0.40;
 pub const SCALE_MAX: f64 = 1.00;
 
-/// One half of the tray icon: which provider, and which of its windows.
-/// `window` is a window id as the provider reports it ("session", "weekly_all", "primary"…), or
-/// the empty string / "top" meaning "whichever of its windows is fullest" — the same rule the
-/// notch ring uses, and the only choice that keeps working when a provider changes its windows.
+/// One half of the tray icon, or one ring on the notch: which provider. It shows that provider's
+/// ring, so the tray and the notch can never disagree. (A `window` key from older builds is ignored.)
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct TraySlot {
     pub provider: String,
-    #[serde(default)]
-    pub window: String,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -60,11 +56,15 @@ pub struct Config {
     /// kept so an existing config migrates cleanly.
     #[serde(default)]
     pub notch_providers: Vec<String>,
-    /// What each ring on the notch shows: the provider, and which of its windows. An empty list
-    /// means every provider, each showing whichever of its windows is fullest — the original
-    /// behaviour. Same shape as the tray slots so the two settings read alike.
+    /// Which providers get a ring on the notch, in order. An empty list means every provider.
     #[serde(default)]
     pub notch_slots: Vec<TraySlot>,
+    /// Antigravity's lane on the ring, as the Mac app's "Notch reads": "automatic", "5h" or "weekly"
+    #[serde(default = "default_antigravity_limit")]
+    pub antigravity_limit: String,
+    /// The model family that choice looks at, as the Mac app's "Model data": "gemini" or "3p"
+    #[serde(default = "default_antigravity_model")]
+    pub antigravity_model: String,
     /// false = the pill is kept off the screen edge entirely; the tray icon is then the only way in
     #[serde(default = "yes")]
     pub notch_visible: bool,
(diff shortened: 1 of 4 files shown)


vinzdg:
This needs a rebase onto `main` before I can take it, and it isn't a trivial one — sorry.

#141 (Russian) landed today and rewrote the Windows UI files. It added a `ui(key, fallback)` helper, an `RU_STATIC` table and `applyStaticUi()` to `settings.html` and `notch.html`, and wrapped 29 JavaScript-built strings in `settings.html` alone. Your branch predates all of it, so the merge conflicts across those files and your side has no `ui()` function at all.

I tried resolving it here: taking your side and re-wrapping the six strings that survive your change gives calls to a function that no longer exists in the file, which would break the settings page at load. Stitching the two versions together by hand is not something I'm willing to do blind, since I have no Rust toolchain or Windows machine to run it on.

Your change itself looks right and deliberate — removing the per-ring window list so Windows follows the Mac's `headlineID` is exactly the parity the Mac app settled on, and the reasoning in the description is sound. It just needs to be applied on top of the localized files, by someone who can run `cargo test` and see the page.

For reference, the Windows CI job from #147 is now on `main`, so a rebased branch will get `cargo build`, `cargo test` and `clippy` run on every push.

RawJat:
Rebased onto main (8c29d29) and force-pushed — thanks for looking at it rather than
closing it.

You were right not to stitch it by hand. The resolution keeps the localized files as
the base and replays the change on top, so ui(), RU_STATIC and applyStaticUi() are all
still there and every ui() call resolves:

- The per-ring window lists this PR removes took their dictionary entries with them,
  including "Whichever is fullest" and the two slot-repair notes.
- The strings that survive, plus the two Antigravity pickers this PR adds, go through
  ui(). Switching to Russian calls render(), so those pickers rebuild in the new
  language rather than going stale.
- The Russian is copied from Sources/Localizable.xcstrings, not invented: "Notch reads"
  is "Показания панели", and the lane and model names were already translated there.
  "Model data" has no entry in that catalog, so it stays English here exactly as it
  does on the Mac.

One thing worth flagging beyond the conflict: two captions this PR shortens had kept
their old RU_STATIC keys, so they silently fell back to English — applyStaticUi matches
text exactly, and nothing errors or fails when a reworded string loses its key. Both
keys now move with the text. For "Click a name to switch its ring on or off." that is
your sentence minus the half describing the window list this PR removes. For "Pick the
tool; each part draws that tool's ring." there was no existing sentence to reuse, so
I assembled "Выберите инструмент: каждая часть показывает его кольцо." from wording
already in the file — @unchase, a second opinion on that one line would be welcome.

Only rendering the page catches this class of bug, so I checked it that way: settings.html
with the bridge stubbed and the language forced to Russian.

Correction: clippy is 28 warnings, not 30 — I'd been counting cargo's two summary
lines. Still identical to main, so neither PR adds one.

PR #153, opened 2026-09-11
 Make the Windows hover card read like the Mac's

RawJat:
Two things on the Windows hover card didn't match the Mac, and both made it harder to read.

**Antigravity's lanes weren't told apart.** Antigravity reports a 5-hour and a weekly limit for Gemini models, and the same pair for Claude and GPT models. The card listed all four flat, each labelled with its group, so it read "Gemini Models" twice with nothing to say which lane was which. Now, as on the Mac:
- each model group is a heading, with its lanes in one box
- lanes are named **5-hour Limit** and **Weekly Limit**, 5-hour first
- both sources get this: the language server (`gemini-5h` IDs) and the `agy` CLI ("Gemini Models Five Hour Limit")

`LimitWindow` gains an optional `group`. A window without one draws as before, so the other providers don't change. The per-window lists in Settings (from #115) now show lane names without their group; #152 removes those lists.

Rebased on `main` after #141 landed Russian. The group heading and the lane names go through its `textCopy()`, and its dictionary gains the three labels this PR puts on the card — "5-hour Limit" and the two model group names — in the Mac catalog's own wording, so a Russian card reads «Лимит на 5 часов» under «Модели Gemini» rather than falling back to English.

**The tail never showed.** The card had a triangle meant to point at the hovered ring. It sat just outside the card, and the card's `overflow:hidden` clipped it. It's now an element of its own:
- drawn as the Mac's curved `TooltipTail` at the Windows card's scale, running from the card to the pill
- its tip on the hovered ring, not the middle of the cell and its label
- reported as part of the hot area, so crossing from the pill to the card along it stays hot at every step

The notch window grows from 340 × 460 to 360 × 520 logical pixels:
- **wider**, so the card keeps its 246 px width with the tail beside it. `fitZoom` divides by the same width.
- **taller**, because two groups plus a stale line and an activity row cut off the card's last row.

The extra area is transparent and click-through.

## Test plan

- [x] `cargo test`: 27 pass.
  - New: the language server's reply is grouped, named and ordered.
  - Updated: `agy_cli`'s parser tests, for the new names and order.
  - Rewritten: the click-through test, which now pins the tail layout.
- [x] `cargo clippy --all-targets`: the same 30 warnings as `main`.
- [x] Rendered `notch.html` in headless Edge with the bridge stubbed: grouped boxes, the tip on the hovered ring, and the worst case (stale plus working) fits.
- [x] The same render with the language set to Russian: «Модели Gemini» and «Модели Claude и GPT» over «Лимит на 5 часов» and «Недельный лимит».
- [x] Live on Windows 11 at 150 %: the grouped Antigravity card.
- [x] Live: the tail, hovering each ring and moving from the pill onto the card.

### DIFF
--- a/windows/codenotch/src/agy_cli.rs
+++ b/windows/codenotch/src/agy_cli.rs
@@ -129,9 +129,16 @@ fn parse_quota(text: &str) -> Result<Vec<LimitWindow>, String> {
             .replace(" and ", "/")
             .replace(" Weekly Limit", " · Weekly")
             .replace(" Five Hour Limit", " · 5h");
+        // Grouped by model family and named by lane, as the Mac card shows them
+        let group = [" Weekly Limit", " Five Hour Limit"]
+            .iter()
+            .find_map(|s| label.strip_suffix(s))
+            .map(String::from);
+        let lane = crate::antigravity::lane_name(&label).filter(|_| group.is_some());
         out.push(LimitWindow {
+            label: lane.map_or(short_label, String::from),
+            group,
             id: label,
-            label: short_label,
             used: ((100.0 - remaining) / 100.0).clamp(0.0, 1.0),
             resets_at: Some(reset as u64),
             ..Default::default()
@@ -140,6 +147,7 @@ fn parse_quota(text: &str) -> Result<Vec<LimitWindow>, String> {
     if out.is_empty() {
         return Err("CLI returned no recognised quota windows".into());
     }
+    crate::antigravity::order_lanes(&mut out);
     Ok(out)
 }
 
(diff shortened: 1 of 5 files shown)


vinzdg:
This needs a rebase onto `main` before I can take it, and it isn't a trivial one — sorry.

#141 (Russian) landed today and rewrote the Windows UI files. It added a `ui(key, fallback)` helper, an `RU_STATIC` table and `applyStaticUi()` to `settings.html` and `notch.html`, and wrapped 29 JavaScript-built strings in `settings.html` alone. Your branch predates all of it, so the merge conflicts across those files and your side has no `ui()` function at all.

I tried resolving it here: taking your side and re-wrapping the six strings that survive your change gives calls to a function that no longer exists in the file, which would break the settings page at load. Stitching the two versions together by hand is not something I'm willing to do blind, since I have no Rust toolchain or Windows machine to run it on.

Your change itself looks right and deliberate — removing the per-ring window list so Windows follows the Mac's `headlineID` is exactly the parity the Mac app settled on, and the reasoning in the description is sound. It just needs to be applied on top of the localized files, by someone who can run `cargo test` and see the page.

For reference, the Windows CI job from #147 is now on `main`, so a rebased branch will get `cargo build`, `cargo test` and `clippy` run on every push.

RawJat:
Rebased onto main (8c29d29) and force-pushed. The conflict here was with #141 in
notch.html, in renderCard's window loop.

The group heading and the lane names now go through its textCopy(), and its dictionary
gains the three labels this PR puts on the card — "5-hour Limit" and the two model
group names — taken from Sources/Localizable.xcstrings. A Russian card reads
"Лимит на 5 часов" under "Модели Gemini" instead of falling back to English.

Correction: clippy is 28 warnings, not 30 — I'd been counting cargo's two summary
lines. Still identical to main, so neither PR adds one.

PR #155, opened 2026-09-11
Connect a phone (Android or iphone) to Codenotch

mohammad-albarham:
Adds pairing so the Codenotch phone app can read usage from the desktop app directly, over the local network, instead of only from the Python agent.

**What's in it**

- `docs/phone-link-protocol.md` — the v2 contract: a one-time 5-minute code in
  the QR/link, a per-device secret derived on both sides and never transmitted,
  and HMAC-signed requests carrying `X-CN-Device`.
  
- `Sources/PhoneLink/` — server, pairing, registry, snapshot builder, and the
  pairing window with the QR code.
  
- A Settings section to pair, list and revoke devices.

- `Tests/PhoneLinkTests.swift` — 307 lines covering pairing, signing, revocation
  and snapshot encoding.

The server binds the local network only and stays off until you open the pairing
window. `NSLocalNetworkUsageDescription` is declared in `project.yml`.


vinzdg:
This is careful work and the security design is better than I expected — 128-bit pairing codes from `SecRandomCopyBytes` with a 5-minute expiry, HMAC-SHA256 on every request compared in constant time, nonce replay rejection, per-IP rate limiting on pairing, a 64 KB body cap, and off by default. Thank you for that.

Before it can be merged, three things are the maintainer's call rather than mine, and I'd like your thinking on them:

1. **It binds `0.0.0.0`** (`PhoneLinkServer.swift:67`), so the listener is on every interface — café and office Wi-Fi included, not just the Mac. Was a narrower bind, or an explicit "only while pairing" window, considered?
2. **It's plain HTTP.** The HMAC authenticates each request but encrypts nothing, so the usage payload crosses the network readable, as does the device id. Over a shared network that is someone's account usage in the clear.
3. **Paired devices' secrets are stored unencrypted** in `devices.json` (`PhoneLinkRegistry.swift:34`, plain `Data.write`). Every other borrowed credential in this app lives in the keychain or belongs to the tool that owns it; this one is a secret Codenotch itself mints, so the keychain seems like its home.

None of these are objections to the feature — it's a genuinely nice idea, and the request signing is the hard part and it's done properly.

mohammad-albarham:
I agree with you, now those should be resolved @vinzdg

vinzdg:
Thanks again for this, @mohammad-albarham. The pairing code is still in main, but I've hidden the Phone pane and the Connect Phone… menu item for now (`PhoneLink.isAvailable = false` in `PhoneLinkServer.swift`), since there's no phone app people can install yet. Once the app is available, point me to it and I'll turn pairing back on.

mohammad-albarham:
Thank you @vinzdg.

Already I have prepared and released the apps, you can find them in the following repo:

https://github.com/mohammad-albarham/codenotch-mobile/releases

I have tested the android app both of them in simulators and I tested android app in my android phone.


You can point the users to this repo:

https://github.com/mohammad-albarham/codenotch-mobile

PR #197, opened 2026-09-14
fix(notch): keep always-on notch visible when full-screen fold is off

mulhamna:
### Summary
- Fix an infinite folding loop in full-screen apps when "Fold for full-screen apps" is disabled with "Always show" enabled.
- Ensures the notch stays stably visible over full-screen apps when full-screen folding is turned off.

### Changes
- **NotchWindowController**: Update `cursorMoved()` to respect `foldsForFullScreen` when determining `ignoreAlwaysOn` during cursor tracking.
- **FullScreenAutoFoldTests**: Add unit tests verifying cursor tracking leaves the always-on notch open and schedules no fold work when auto-fold is off.

### Test plan
- [x] Run `FullScreenAutoFoldTests` (12/12 passing).
- [x] Verify always-on notch remains open during cursor moves under full-screen apps when auto-fold is disabled.
- [x] Verify fold work is still scheduled when auto-fold is enabled.


vinzdg:
Thanks @mulhamna, you found the right line. #194 made the same fix in `cursorMoved` and also cancels a fold that's already scheduled when the setting is switched off (otherwise the notch can stay folded), with tests that wait for the fold to land. That's merged now, so I'm closing this as a duplicate. Much appreciated.

PR #202, opened 2026-09-15
fix: optimize cursor tracking and full-screen polling to eliminate battery drain

### LINKED ISSUE #185: fix: choppy notch hover animations and high CPU on macOS (1.11.0)
### Bug Description

Hovering over the notch produces visibly choppy animations on Codenotch 1.11.0. CPU observations during investigation were approximately 44–99% for the Codenotch process (100% represents one CPU core).

A five-second sample of the installed app points to substantial main-thread time spent loading/parsing localization strings. This is a profiling lead, not a confirmed root cause or a verified fix.

### Steps to Reproduce

1. Run Codenotch 1.11.0 with provider rings visible.
2. The observed configuration uses the right screen edge, **Always show**, medium size, and **Liquid Glass**. The app language follows the system (Simplified Chinese).
3. Move the pointer over the notch and observe its hover animations: they are visibly choppy on this machine.
4. Inspect the Codenotch process in Activity Monitor. To investigate the same hot path, capture a sample while the problem is present:

   ```sh
   sample "$(pgrep -x Codenotch)" 5 -file codenotch-sample.txt
   ```

These steps describe the affected setup; I have not established the smallest required combination of providers/settings. I have not measured animation frame times or recorded a synchronized hover trace.

### Expected Behavior

Hover animations should remain smooth, with modest CPU use for a usage-monitoring UI.

### Affected Provider (if applicable)

UI / Notch Presentation

### Environment & macOS Version

- macOS 26.6.2 (25G83)
- Apple M4 Pro, 14 logical CPU cores, 48 GB RAM
- System locale: `zh_CN`; app language: follow system
- During CPU observations, overall system CPU idle was approximately 64–76%; memory pressure was normal (`kern.memorystatus_vm_pressure_level = 1`).

### Codenotch Version

Installed release **1.11.0 (13)**. Source references below are pinned to release commit `6c28672c300fadcb7fb59a277ba9d3997d74f1d2`.

### Relevant Logs (Unified Log)

CPU profiling was used rather than Unified Log for this report.

- One Activity Monitor capture showed **Codenotch: 73.8% CPU** and **system idle: 76.11%**.
- `sample` ran for 5 seconds, requesting a 1 ms sampling interval.
- The main thread had **3,462 samples**. Summing its `_loadStringsFromData` call-tree entries gives **2,209 samples (63.8%)**. This is a share of main-thread samples, not a percentage of total machine CPU.
- A representative observed localization stack is:

  ```text
  Codenotch [app frames are stripped]
    String.init(localized:)
      String.init(localized:options:)
        LocalizedStringResource._withExtensionForBundle
          NSBundle._localizedStringForKey:value:table:localizations:
            _CFBundleCopyLocalizedStringForLocalizationTableURLAndMarkdownOption
              _copyStringTable
                _loadStringsFromData
                  __CFBundleCreateStringsFromPlistData
                    _CFPropertyListCreateFiltered
  ```

The sample includes timer and event-handling paths. Because app symbols are stripped, the exact application-level caller still needs confirmation in Instruments or a symbolicated build.

#### Possible hot path to investigate

- [`L10n.t`](https://github.com/vinzdg/codenotch/blob/6c28672c300fadcb7fb59a277ba9d3997d74f1d2/Sources/L10n.swift#L60-L68) performs each lookup using `LocalizedStringResource(... bundle: .atURL(bundle.bundleURL))`.
- [`startWatchingCursor`](https://github.com/vinzdg/codenotch/blob/6c28672c300fadcb7fb59a277ba9d3997d74f1d2/Sources/Notch/NotchWindowController.swift#L510-L532) invokes cursor handling on a 0.3-second timer and on mouse-move events. [`cursorMoved`](https://github.com/vinzdg/codenotch/blob/6c28672c300fadcb7fb59a277ba9d3997d74f1d2/Sources/Notch/NotchWindowController.swift#L540-L601) performs repeated geometry/hit-region calculations.
- [`contentCardHeight` / `maxCardHeight`](https://github.com/vinzdg/codenotch/blob/6c28672c300fadcb7fb59a277ba9d3997d74f1d2/Sources/Notch/NotchViewModel.swift#L576-L603) recompute card heights across snapshots, including status/block text.

These are candidate contributors based on the profile plus source inspection. Caching localization work and unchanged geometry may be worth investigating, while preserving runtime language changes and interpolated strings.

#### Small localization comparison using the installed app's resources

This standalone Swift probe repeats the same fixed key (`Thinking`) 500 times. It compares the resource lookup used by `L10n.t` with a reused language-specific `Bundle`. The latter is only a comparison for a fixed key, **not a proposed drop-in replacement** for the full localization API.

| Lookup | Locale | 500 calls, total | Per call |
| --- | --- | ---: | ---: |
| `LocalizedStringResource` + `.atURL` | `en` | 168.72 ms | 0.337 ms |
| `LocalizedStringResource` + `.atURL` | `zh-Hans` | 164.56 ms | 0.329 ms |
| `LocalizedStringResource` + `.atURL` | `zh_CN` | 159.25 ms | 0.318 ms |
| Reused `Bundle.localizedString` | `zh-Hans` | 1.37 ms | 0.003 ms |

The Chinese lookups returned the same total character count in both paths. The English resource lookup was similarly expensive, so this does not establish a Chinese-only issue. These are single-run microbenchmark results, not a full-app before/after comparison.

<details>
<summary>Runnable Swift probe</summary>

Save as `localization-probe.swift` and run `swift localization-probe.swift` with Codenotch installed at `/Applications/Codenotch.app`.

```swift
import Foundation

let appURL = URL(fileURLWithPath: "/Applications/Codenotch.app")
let key: String.LocalizationValue = "Thinking"
let iterations = 500

for localeID in ["en", "zh-Hans", Locale.current.identifier] {
    let locale = Locale(identifier: localeID)
    var length = 0
    let start = Date()
    for _ in 0..<iterations {
        length += String(localized: LocalizedStringResource(
            key, locale: locale, bundle: .atURL(appURL)
        )).count
    }
    let elapsed = Date().timeIntervalSince(start) * 1000
    print("Codenotch lookup locale=\(localeID) n=\(iterations) total_ms=\(String(format: "%.2f", elapsed)) per_call_ms=\(String(format: "%.3f", elapsed / Double(iterations))) result_chars=\(length)")
}

if let bundle = Bundle(path: "/Applications/Codenotch.app/Contents/Resources/zh-Hans.lproj") {
    var length = 0
    let start = Date()
    for _ in 0..<iterations {
        length += bundle.localizedString(forKey: "Thinking", value: nil, table: nil).count
    }
    let elapsed = Date().timeIntervalSince(start) * 1000
    print("Cached Bundle lookup locale=zh-Hans n=\(iterations) total_ms=\(String(format: "%.2f", elapsed)) per_call_ms=\(String(format: "%.3f", elapsed / Double(iterations))) result_chars=\(length)")
}
```

</details>

No application patch has been applied, and no change to language, surface style, or geometry caching has been tested end to end. Happy to help test a targeted fix.

divingsbysangam:
## Summary

Addresses Issue #185 and high battery/CPU consumption during normal pointer movements and system idle. (Supersedes closed #201).

## Root Cause

1. **Eager Evaluation of `isFullScreenActive()`**: In `cursorMoved()`, `setExpanded(..., ignoreAlwaysOn: foldsForFullScreen && isFullScreenActive())` eagerly evaluated its arguments. Because global monitors for `.mouseMoved` and `.leftMouseDragged` invoke `cursorMoved()` for every mouse movement anywhere on the system (60–120Hz), `FullScreenDetector.isFullScreenAppFrontmost()` called `CGWindowListCopyWindowInfo([.optionOnScreenOnly, .excludeDesktopElements], kCGNullWindowID)` dozens to hundreds of times per second. This caused heavy Mach IPC traffic to WindowServer, dictionary allocations, and 44–99% CPU usage.
2. **Missing Fast-Path**: When the notch is folded shut and the cursor is elsewhere on the screen, `cursorMoved()` ran full layout, tooltip, and interactive rect checks on every mouse move.
3. **Unnecessary 0.3s Recurring Polling**: `cursorTimer` repeatedly ran `handleActiveSpaceOrAppChange()` every 300ms 24/7, keeping CPU cores awake.
4. **Redundant WindowServer IPC**: `updateInteractiveRects()` set `panel.ignoresMouseEvents` without checking if the boolean value had changed.

## Key Changes

- **Gated Evaluation**: Changed `ignoreAlwaysOn` in `setExpanded()` to `@autoclosure () -> Bool = false`. It is only evaluated when `!wanted && model.isExpanded && model.isAlwaysOn && !model.isPinned && foldsForFullScreen`. For normal mouse movements, `isFullScreenActive()` and `CGWindowListCopyWindowInfo` are never invoked.
- **Pointer Fast Path**: In `cursorMoved()`, if `!model.isExpanded && !panel.frame.contains(NSEvent.mouseLocation)`, return immediately.
- **Event-Driven Space Changes**: Removed `handleActiveSpaceOrAppChange()` from the recurring cursor timer. Space transitions are handled upon `activeSpaceDidChangeNotification` and `didActivateApplicationNotification`, with a debounced `+0.35s` follow-up to cover space slide animations.
- **Coalesced Timer**: Added `poll.tolerance = 0.1` and idle pointer suppression when the cursor has not moved and the notch is folded shut.
- **WindowServer Caching**: Added 0.5s TTL caching and `invalidateCache()` to `FullScreenDetector.isFullScreenAppFrontmost(on:)`.
- **Property Guard**: Guarded `panel.ignoresMouseEvents` assignment in `updateInteractiveRects()` so it only writes when changed.

## Verification

- `FullScreenAutoFoldTests` (14/14 tests pass, including new test `testCursorMovedSkipsFullScreenCheckWhenFoldedAndCursorOutside`).
- All full-screen folding, auto-restoration, hover expansions, and pinning behaviors verified.

Closes #185


vinzdg:
Thanks @divingsbysangam — the battery goal is right and the `@autoclosure` on `setExpanded` is a good catch, but I can't take this as it stands: the poll no longer calls `handleActiveSpaceOrAppChange()`, so full-screen state is only re-evaluated on `activeSpaceDidChange`/`didActivateApplication`. Non-native full screen posts neither — a video going full screen in the browser that's already frontmost, or a game that sizes a window to the display — so the notch would never fold for it, and the always-on restore on exit is lost too. That's the case the file exists for.

Could you keep a (slower) poll, or drive it from a display/window notification that covers those cases, and make the test assert a real fold/unfold transition? `testCursorMovedSkipsFullScreenCheckWhenFoldedAndCursorOutside` currently asserts inside `if let frame, !frame.contains(NSEvent.mouseLocation)`, so it passes vacuously depending on where the pointer happens to be.

divingsbysangam:
Thanks for the clear feedback @vinzdg! I have pushed an update addressing both points:

1. **Slower full-screen poll**: Added a dedicated `fullScreenTimer` polling at `2.0s` (with `0.5s` tolerance) in `NotchWindowController`. This preserves timer coalescing and low-power core sleep while reliably detecting non-native full-screen state (browser video / game windows) and restoring always-on notches on exit.
2. **Injected pointer provider & hardened tests**: Added `mouseLocation: () -> CGPoint` to `NotchWindowController` so tests are deterministic. Fixed `testCursorMovedSkipsFullScreenCheckWhenFoldedAndCursorOutside` to remove the vacuous pointer conditional, and added `testFullScreenPollFoldsWhenNonNativeFullScreenActivatesAndRestoresWhenExiting` to verify both fold and unfold transitions driven by the periodic poll.

vinzdg:
Thanks @divingsbysangam — your update fixed the non-native full-screen case. Since then main landed a CPU fix that already covers the ignoresMouseEvents dedupe, the lazy full-screen check in `setExpanded` and a cached full-screen reading, so most of this PR is now on main. The piece main didn't have was yours: a separate 2-second full-screen timer instead of riding the 0.3 s cursor poll, plus the follow-up check 0.35 s after a space or app switch. I ported that in 713e31e, credited to you. Closing this one — thank you, it made a real difference.

PR #207, opened 2026-09-15
perf: stop re-reading megabytes per tick in the activity monitors

Petagorus:
## Summary

On machines with large agent histories, Codenotch saturates a CPU core
permanently (~96% measured) and the notch UI visibly drops frames. `sample`
shows the main thread spending essentially 100% of its time inside
`__NSFireTimer` callbacks doing per-line JSON parsing and directory scans —
the two-second tick re-reads far more data than it needs.

This PR bounds every expensive read I found on that path. Idle CPU on a
machine with ~13 GB of agent transcripts drops from ~96% to ~3%.

## Root causes (all measured, not guessed)

- **Whole-rollout parse every 2 s.** `CodexActivityMonitor` called
  `CodexRolloutActivity.state(from:)` on the newest rollout on every tick.
  `state` did `Data(contentsOf:)` + `String.split` + `JSONSerialization` on
  *every line*. A live rollout here is **292 MB** — each tick took longer
  than the interval, so the timer ran back-to-back. It also ran for stale
  rollouts whose session `session(...)` then discarded on the same
  `staleAfter` check — pure waste.
- **Full SQLite scan every 2 s.** `CodexStore.newestRollout` /
  `newestDesktopThread` open and scan `state_5.sqlite` — **427 MB** on this
  machine, WAL-mode — once each per tick.
- **Per-conversation attribute reads.** `AntigravityActivityMonitor`
  stat-ed `.system_generated/logs/transcript.jsonl` inside every
  conversation directory ever written (~1,700 here) using
  `attributesOfItem` — which pulls all extended attributes via getxattr
  chains — plus `appendingPathComponent`, which itself syscalls per call.

## Changes

- `CodexRolloutActivity.state(from:)`: scans the rollout **backwards in
  256 KB windows** for the newest lifecycle event, capped at 4 windows
  (1 MB). Semantics preserved: the newest event still wins, a window's
  cut first line is carried into the earlier window rather than parsed
  half, and a miss maps to `.busy` exactly as before — a fresh file with
  no lifecycle event in its last megabyte is a turn in progress in every
  realistic case.
- `CodexRolloutActivity.state(from:)`: reads now loop until each 256 KB
  window is filled — `read(upToCount:)` may legally return short, which
  would have silently lost lines and mis-joined the carried fragment.
- `CodexActivityMonitor.read`: the `staleAfter` freshness check now runs
  **before** `state(from:)` — a stale rollout is never parsed, since
  `session` would have dropped it anyway.
- New `CodexStoreCache`: stat-gates both store lookups on `(mtime, size)`
  of the database **and its `-wal` file** (writes land in the wal first;
  the main file's mtime only moves at checkpoint). An unchanged store now
  costs a few stats instead of a SQLite open + scan. Wired through
  `read(..., cache:)` with a default instance so existing callers/tests
  are untouched.
- `AntigravityActivityMonitor.read(root:)`: one `stat()` per transcript on
  a plain string path — no `attributesOfItem` xattr reads, no
  `appendingPathComponent` filesystem calls inside the loop. The winner's
  URL is constructed once, after the loop.

**Deliberate trade-off:** a `task_complete` further than 1 MB from the
rollout's end is never seen — that turn reads as still-busy until the
rollout goes stale, and its completion chime is skipped. Bounding the
scan is the point of the change (a mid-turn `task_started` can sit
arbitrarily far behind the writes streaming in), and in the common case
a file that grew past the cap has another turn already underway, where
busy is the right answer anyway.

## Verification

Measured on macOS 26.4, ~13 GB of agent transcripts
(`~/.codex` 16 GB, `~/.claude/projects` 7 GB, `~/.gemini` 3.7 GB):

| | before | after |
|---|---|---|
| Idle CPU (`ps %cpu`) | ~96% | ~2–3% |
| Main-thread time inside timer callbacks (`sample`) | ~100% | minority, spread across all monitors |

- `make test-ci`: 1533/1535 pass. The 2 failures —
  `KiroUsageTests.testAnISOResetDateIsReadAsPrinted` and
  `testReadsTheTwoDotXEstimatedUsageCard` — a `dd.mm` year-parse
  bug in the Kiro provider, whose code this diff does not touch.
- All 6 `CodexActivityTests` cases pass, including the three
  `CodexRolloutActivity.state` lifecycle cases.

## Test plan

- [x] Unit tests for rollout lifecycle events (`CodexActivityTests`)
- [x] Ran the built app against real stores; sampled CPU before/after
- [ ] Maintainer re-checks on a machine with a much smaller
      `~/.codex` — the caches are stat-gated, so small stores are a
      subset of the tested path


vinzdg:
Thanks @Petagorus — the shape of this is right, and the backwards window scan handles truncation, rotation and UTF-8 boundaries correctly as far as I can tell. Two things before I merge:

1. **Short reads.** `handle.read(upToCount:)` may legally return fewer bytes than asked for. When it does, the window silently loses lines and the carried fragment is mis-joined. Please loop until the window is filled (or EOF).
2. **No tests.** For a rewrite of this shape I'd like coverage of: the backwards window splitting, the carried partial first line, a file smaller than one window, and cache invalidation when mtime/size change.

Also worth a note in the PR: a `task_complete` further than 1 MB from the end now reads as still-busy, so that turn's completion chime is skipped. Fine by me, but it should be deliberate.

The full suite passes locally on this branch (1,535 tests).

Petagorus:
Done — `2bd9c5e`.

- **Short reads:** the window fill now loops on `read(upToCount:)` until the window is complete or EOF. A read error still fails the scan; an early EOF means the file shrank between seek and read, and the partial window is still contiguous with `windowStart` — so it is scanned rather than dropped. `carried` is only joined when the window actually reached `windowEnd`, so a short window can't fabricate a line.
- **Tests added** (11): earlier-window events are still found, the boundary-cut lifecycle line is reassembled, a file smaller than one window reads whole, a half-written tail line doesn't hide the event behind it, an event beyond the 4-window cap is not found, and the store cache re-scans on mtime change, size change alone, `-wal` change alone, and a deleted cached path — plus the same stamp gate for the desktop catalogue.
- **PR body:** noted the 1 MB cap as deliberate — a `task_complete` further than that from the end reads as still-busy and its chime is skipped.

`make test-ci`: 1544/1546 pass; the two `KiroUsageTests` failures are the pre-existing `dd.mm` year-parse bug this diff doesn't touch.

PR #210, opened 2026-09-15
fix(windows): select core Codex quota and recover native-client reads

rustamnova:
## Summary

Fix Windows Codex quota selection and add a native-client recovery path when the direct usage read fails. Updated after review and rebased onto `main` at `4462432` (1.12.0) on September 16.

## Why

- Codex rollout logs may contain separate core and Spark limit events. Taking the last event without checking its bucket can show Spark in place of the core allowance.
- If core windows are absent but extra windows exist, `windows.first()` can promote Spark or code review to the main ring/tray.
- Resuming an old Codex thread modifies its original rollout file; searching only three recent creation-date directories misses it.
- The direct HTTP adapter cannot always use the same sign-in path as an installed Codex client. A failed direct read should have a bounded native-client recovery option before showing a stale rollout.

## Changes

- Keep the existing HTTP endpoint first and preserve five-minute polling and HTTP 429 backoff.
- On failure outside that backoff, use an installed native `codex.exe` and the documented [`account/rateLimits/read`](https://learn.chatgpt.com/docs/app-server#6-rate-limits-chatgpt) app-server method. Discover the Windows desktop binary as well as existing native CLI candidates; never launch a `.cmd`/Node wrapper. Hide the owned process, bound the response wait to 20 seconds, and kill/reap that process after the read. Send only initialization and the quota read, not login or inference requests.
- Prefer `rateLimitsByLimitId.codex` over the legacy single-bucket reply. Accept legacy core replies, reject explicitly different buckets, and do not turn missing usage into zero.
- Select **only `primary`** in both the Rust tray and JavaScript headline, matching macOS. Missing primary stays blank; `secondary` remains available to the separate weekly ring. Keep additional limits as separate details on the existing HTTP path.
- Filter non-core rollout events. Like macOS, use `state_5.sqlite` to find recently updated non-archived threads, checking at most eight candidate paths. The connection is read-only and WAL-aware, with a 50 ms busy timeout; both millisecond and legacy second timestamp columns are supported. If the index cannot be read, keep the original three-date-directory fallback, not a scan of all session files.
- Add parser/filesystem/tray regression tests and dependency-free Node tests against explicitly marked production selectors (no dependency on neighbouring function names/order). Cover the separate weekly ring and run these tests in Windows CI.

## Why retain native recovery?

The motivating local case was a failed stored-token HTTP read while the installed native Codex client could obtain the quota. Codex owns its managed OAuth lifecycle, avoiding duplicated refresh handling in Codenotch. This is **not** a return to unconditional wrapper spawning: HTTP still wins, 429 backoff is respected, no `.cmd`/Node tree or `taskkill` is used, and the owned native process is bounded and reaped. Codenotch sends no login or explicit token-refresh request; Codex may perform its own normal managed refresh.

This is not guaranteed to work with external credentials that need host-app support. Failure still falls back to an honestly stale/missing rollout. The behavior and multi-bucket schema are documented in [OpenAI's app-server documentation](https://learn.chatgpt.com/docs/app-server#6-rate-limits-chatgpt).

## Scope and tradeoffs

- This preserves the upstream **used percentage** convention and the generic **Codex** label. It reports quota utilization, not exact tokens or a separate allowance for a particular model.
- Local Astra branding, remaining-percentage UI, Claude/proxy changes, personal configuration, credentials, and logs are intentionally excluded.
- The rollout tail remains 256 KiB and activity polling is unchanged. The all-history metadata scan has been removed; old resumed threads require a readable thread index to be discovered outside the bounded directory fallback.
- macOS code and other provider adapters are unchanged. The installed local widget was not replaced by this branch.

## Validation

- `cargo test --release --locked -j 6`: **80 passed, 0 failed, 3 ignored** (opt-in live-account integration checks).
- `node --test test-codex-headline.cjs`: **5 passed, 0 failed**; also parses all inline notch scripts for syntax errors.
- `cargo test --release --locked codex::tests::live_native_quota -- --ignored`: **1 passed**, explicitly checking this branch's native transport against an already signed-in local client; no credentials or quota values printed/published.
- `git diff --check`: passed.
- `cargo clippy --release --all-targets --locked -j 6`: passed (exit 0, no warnings emitted).
- `cargo build --release --locked -j 6`: passed; the installed widget was not replaced.
- macOS `make test` was not run on this Windows machine. No private account fixtures are included.

GitHub Actions is gated on **maintainer approval**, not a reported test failure. The refreshed [Windows run for `99c3a8b`](https://github.com/vinzdg/codenotch/actions/runs/35093782473) explicitly says it is awaiting approval and has no started jobs. The [original Windows run](https://github.com/vinzdg/codenotch/actions/runs/34977768311) was gated too. Local results above are separate from GitHub CI results.


vinzdg:
Thanks @rustamnova. Two things before this can go in:

1. **No CI has run on this PR at all** — every check is missing, so nothing here has been built or tested. Could you push a rebase on current main so the Windows job runs? Note main has moved a lot today (#208/#209/#213 rebuilt the settings window, added notch sizes and the weekly ring; #212 recoloured the arc; #183 rewrote the card's language tables), so `notch.html` and `main.rs` will need a look.
2. **Codex parity:** `.or_else(|| by_id("secondary"))` puts the weekly number where the session window belongs. The Mac deliberately leaves that blank instead — see `CodexLocalProvider` setting only `headlineID: "primary"`, and the comment in `UsageModel` that a blank is honest where a weekly percentage wearing the session's place is not. Please drop the fallback.

Smaller points: `newest_rollout` lost its 3-day cap, so it now stats every session file every 5 minutes; and `read_app_server` brings back process spawning that the module docstring says was removed in 1.5.0 — bounded and reaped, but tell me why it's needed. `test-codex-headline.cjs` extracting source between two function names will break on any reordering; keying it on markers would be sturdier.

rustamnova:
Thanks for the detailed review, @vinzdg. Updated this PR in `99c3a8b`:

1. **Rebased onto current main** (`4462432`, 1.12.0). Reviewed the resulting `notch.html`/`main.rs` delta: the new settings, sizes, weekly ring, colours and translations remain intact. The headline now selects **only `primary`** in both JS and Rust. A missing primary stays blank; secondary is only used by the separate weekly selector.
2. **Removed the all-history file scan.** Like the Mac provider, Windows now reads the latest non-archived paths from `state_5.sqlite`, examining at most eight candidates. Read-only, WAL-aware connection, 50 ms busy timeout; if it cannot read the index, the original three-date-directory fallback remains. Tests cover resumed old threads via committed WAL updates, archived/missing paths, legacy timestamp columns, a corrupt index and the candidate bound.
3. **Replaced function-name slicing with explicit begin/end markers.** Tests cover primary absence, weekly-ring independence and extraction after neighbouring functions are reordered.
4. **Documented why native recovery is retained.** In the motivating Windows case, the stored-token HTTP request failed while the installed native Codex client could read the quota. The client owns its managed OAuth lifecycle, so Codenotch need not implement refresh itself. This is fallback-only (not the old unconditional wrapper path), preserves HTTP 429 backoff, directly launches a hidden native executable, waits at most 20 seconds and kills/reaps its own child. No cmd/Node wrapper, taskkill, login, inference or explicit refresh request. Externally managed auth may still fail; that keeps the ordinary stale/missing fallback. README links the official protocol and explains these limits.

**Local Windows validation:** release build passed; Rust suite **80 passed / 3 opt-in tests ignored**; Node **5 passed**; Clippy release/all-targets passed without warnings. I also explicitly ran the new ignored native-client integration test against an existing local sign-in: **1 passed**. No account data is included in fixtures or output. macOS tests were not run locally.

**CI still needs your approval:** the rebase/push created [Windows run 35093782473](https://github.com/vinzdg/codenotch/actions/runs/35093782473), but GitHub displays “This workflow is awaiting approval from a maintainer” and the jobs list is empty. The original run was gated the same way. Could you approve the refreshed workflows? I cannot approve upstream fork runs with this account, and I have not changed the workflow security policy. These are local test results, not a claim that upstream CI is green.

vinzdg:
Thanks @rustamnova — the headline now matches the Mac (no weekly fallback) and the test markers are solid; the test passes 5/5 on a merge with main. Two things left:

1. **`read_app_server` runs on every failed poll**, including signed-out and rejected-login cases: up to a 20 s blocking `codex.exe` launch every 5 minutes with no cooldown. Please limit it to the one failure it's meant to recover from and add a cooldown, or drop it.
2. **`state_5.sqlite` has no age limit** (`codex.rs` ~382), so it can return a months-old thread that `activity.rs` then reads. Please apply the same 3-day cap the directory walk uses.

Windows CI will run on your next push.

vinzdg:
Merged as `9b6e500`. Thanks for your patience on this one @rustamnova — it sat a while, but the conflict turned out to be a single line: main had moved `staleOf` to the Mac's 15-minute `UsageStore.staleAfter` while your branch added the `END TESTABLE WEEKLY SELECTOR` marker directly above it. Kept both; `test-codex-headline.cjs` passes 5/5 on the merge.

The read-only WAL SQLite with `LIMIT 8`, the killed-and-reaped child and keeping response bodies out of the error text are all right, and the headline now matching the Mac (no weekly fallback) is the part I most wanted.

One follow-up on top, `5c74308`: `read_app_server()` ran after *any* HTTP failure, so a machine with `codex.exe` installed but signed out spawned a hidden process — up to 20 seconds — every five minutes for as long as the app was open. The native path is there for managed sign-ins `auth.json` can't describe, so it's worth one attempt rather than one per poll; a failed read now stands it down for half an hour.

PR #221, opened 2026-09-16
feat(providers): read Claude and Codex usage from browser caches

mulhamna:
## Summary

- Claude and Codex usage tracking currently depends on active CLI logins (`claude /usage`, `~/.codex/auth.json`) or keychain OAuth credentials, which can expire or require repeated authorization prompts.
- Users who work in web browsers (Chrome, Brave, Edge, Arc, Chromium) or ChatGPT Desktop already have their usage API responses stored in Chromium Simple Cache on disk.
- Adding browser cache reading provides a zero-auth, zero-prompt, read-only fallback source that populates live usage without modifying credentials or making background login calls.

## Changes

- **Claude Desktop / Browser Cache** (`Sources/Providers/ClaudeDesktopUsageCache.swift`):
  - Expand candidate cache directories to scan Chromium browser caches (`Google/Chrome`, `BraveSoftware/Brave-Browser`, `Microsoft Edge`, `Arc/User Data`, `Chromium`) in addition to Claude Desktop.
  - Return the freshest reading found across configured directories.
- **Codex Browser Cache** (`Sources/Providers/CodexBrowserUsageCache.swift`, `Sources/Providers/CodexLocalProvider.swift`):
  - Add `CodexBrowserUsageCache` to parse Chromium Simple Cache entries for `chatgpt.com/backend-api/wham/usage`.
  - Fall back to the browser cache snapshot in `CodexLocalProvider` when `~/.codex/auth.json` is missing or expired, or when a local token encounters an authentication error.
- **Tests** (`Tests/CodexBrowserUsageCacheTests.swift`):
  - Add hermetic tests verifying key recognition, response parsing, date extraction, and directory precedence.

## Test plan

- [x] Run `ClaudeDesktopUsageCacheTests`: 38 passed, 0 failed
- [x] Run `CodexBrowserUsageCacheTests`: 7 passed, 0 failed
- [x] Run `CodexUsageTests`, `CodexProfileTests`, `ClaudeOAuthProviderTests`: 57 passed, 0 failed
- [x] Verify project generation (`make gen`) and build integrity with ad-hoc signing


vinzdg:
Thanks @mulhamna — I can see the effort here, but I'm going to decline this one, mainly on privacy grounds:

- **It scans browser caches for everyone, with no opt-in.** Each poll lists up to 400 cache entries per profile per browser (Chrome, Arc, Brave, Edge, Chromium) and reads each entry's header, which holds its URL — in effect reading browsing history on a timer. Codenotch's promise is that it reads only what each tool keeps for itself, so this would need to be off by default and clearly explained, at the very least.
- **It can show the wrong account.** The cached payload isn't matched to the Codex profile, so every profile whose own credentials are missing would show whichever ChatGPT account the browser is signed into, marked ok.
- **It hides sign-in problems.** On 401/403, or with no auth.json, it returns cached numbers marked official instead of needsAuth, so the sign-in prompt stops appearing.
- **Cost.** Listing whole cache folders every poll works against the CPU work that just landed.
- **The Claude half isn't wired:** `ClaudeDesktopUsageCache.candidateDirectories` is never called.

If there's a narrower version you'd like to try — e.g. only the ChatGPT desktop app's own cache, only as a fallback the user switches on, and tied to the right account — I'm happy to look at that.

PR #223, opened 2026-09-16
Fix the empty Windows Settings window left by a merge

unchase:
## Summary

On Windows 1.12.0 the Settings window opens empty: the sidebar shows Accounts, Appearance and General, but every pane is blank and the labels stay in English whatever the language setting.

The settings script does not parse. Merging #208 (Ukrainian localization) kept the body of the old `remoteSwitch()` helper without its opening `function` line, together with the `autoSwitch` / `hooksSwitch` constants that called it. They sit right after the Recentre button handler. The stray `}` is a `SyntaxError`, so none of the page's script runs. The new page already has its own `remote()` helper and `autostart` / `hooks` switches, so the fragment was dead code as well.

`settings.html` still parses at 25cdc9a (before the merge) and fails from 8fb7d8a onwards.

## Changes

- `windows/codenotch/ui/settings.html`: remove the leftover `remoteSwitch` fragment (22 lines). Nothing else changes.

## Test Plan

- [x] Extracted the page script and ran `node --check`: fails on `main` (`SyntaxError: Unexpected token '}'`), passes with this change.
- [x] Opened the page in a browser with a stubbed `window.__TAURI__`: Accounts, Appearance (including the new Size row) and General render in Russian, with no console errors.
- [x] Built with `cargo build --release` on Windows 11 and ran it: the app starts.
- [ ] Settings window opened in the real app: not yet confirmed by eye.
- [ ] macOS: not touched.


RawJat:
Confirmed, and thanks for catching it — #208 is mine, so this one's on me.

My branch had no `remoteSwitch`; `main` just before the merge had three. Neither diff was wrong alone: #219 edited the old page, mine replaced that file wholesale, and git kept the old helper's body without its opening line. Diffing my branch head against `main` shows only #219's two Ukrainian lines plus the 22 you're deleting, and that code is dead anyway — the new page has its own `remote()` helper and switches.

I checked out this branch and ran `node --check` on the page's script: passes. `notch.html` parses, so only the settings window was affected.

Follow-up I'd like to open: the Windows workflow only compiles Rust, so these pages have never been parsed. A `node --check` step would have failed while #208 was still open.

jeffh-flnp:
Confirmed on the real app — this ticks the box you left open.

Environment: Windows 11, built from source with `cargo build --release` (MSVC toolchain), fresh clone of `main` first, then `git fetch origin pull/223/head` and rebuilt.

Before: Settings opened with the Accounts / Appearance / General sidebar and every pane blank, labels stuck in English.
After: all three panes render and the window is interactive.

Thanks — this was the last thing blocking a working install here.

PR #232, opened 2026-09-17
Windows: restore the settings window's script

1boooook2-maker:
The Windows settings window opens with its sidebar drawn and nothing else: an
empty pane, and not one control that answers a click. Quit, Close, the tabs —
none of them.

## Cause

`ui/settings.html` carries the body of a `remoteSwitch` function whose header is
gone. What is left at top level is a bare `return` and one unmatched brace, so
the file does not parse:

```console
$ node --check settings-script.js
SyntaxError: Unexpected token '}'
```

A SyntaxError costs the whole `<script>` block, not just that function. No
handler is ever bound and `boot()` never runs, which is exactly what is on
screen: the panes stay `hidden` as the markup leaves them, and the sidebar does
nothing.

## Fix

The stranded body is also dead code. Its two callers pass `sw-hooks`, an id the
page does not contain, and the switches that actually work are the `remote()`
pair defined above it. So this deletes the block rather than inventing a header
for it. No behaviour is removed.

## Verification

- `node --check` on the extracted script passes.
- Built and run: the window renders and every control responds.
- `cargo test` on `windows/`: 70 passed, 0 failed.

Reproduced on a clean clone of `main` (3251208) with no other patches, so this
is not local to my machine.


vinzdg:
Thanks @1boooook2-maker — you found the same thing: the stray remoteSwitch fragment my conflict resolution left behind, which broke the whole settings script. #223 removed the identical 22 lines a little earlier and is merged now, so I'm closing this one as a duplicate. Appreciated.

PR #234, opened 2026-09-17
Windows: pin the notch to any screen edge, on any monitor

1boooook2-maker:
`place_notch` pinned the notch to the right-hand edge of the primary monitor —
"the other edges are a later milestone", and a second screen was out of reach
entirely. This implements both.

## What changes

**Edge.** Left, right, top or bottom. It stands upright on the left and right
edges, with the hover card opening sideways as before, and lies flat on the top
and bottom ones, with the card opening below or above. The page mirrors or
rotates itself from a `notch_edge` event — Rust owns the edge because Rust is
what placed the window.

The fillets follow one rule on every edge rather than four copies of the
gradient maths: the radial gradient is centred on the corner of the R×R square
*diagonally opposite* the corner where the pill meets the screen, so what stays
black is the concave arc between them.

**Monitor.** Remembered by device name (`\\.\DISPLAY2`), falling back to the
primary one when that screen is no longer attached, so unplugging a display
cannot strand the notch off-screen. Recentre resets the edge and the screen too,
which is what that button is for.

**Dragging does both at once.** Pick the pill up, drop it anywhere: it snaps to
the nearest edge of the screen it was dropped on. The monitor under the window's
centre owns it, and the smallest of the four edge distances wins. No mode to
switch.

**Settings** gains an Edge segment and a Screen list. The list only appears once
there is a second screen, and a monitor the platform will not name is listed
disabled — there would be nothing stable to remember it by.

The window keeps its 360×520 logical size on every edge, so the zoom and DPR
correction are untouched. On the flat edges the card has the pill's thickness
less room, which is the one visible trade-off.

## Verification

Three monitors, two DPI scales (125 % and 200 %):

- each of the four edges places and renders, fillets included;
- the card and its tail follow the hovered ring on all four (tail pointing
  sideways on the upright edges, up or down on the flat ones);
- a drag from the right edge of the 125 % screen to the bottom of the 200 % one
  saves `edge=bottom`, the new monitor and the position along it, and re-places
  with the new scale applied;
- `cargo test` on `windows/`: 70 passed, 0 failed.

## Update: the window is now square

The fillets are drawn outside the pill, so the window has to hold both. A flat
pill is as long as the cells it holds — four cells are 302 px plus 52 of fillet
against a 360 px window, six pixels of slack — and it clipped. Without its
fillets the pill reads as a floating panel rather than something cut into the
edge, which is the point of the shape.

The window is now 560 x 560 on every edge: the long side holds a five-cell pill
with both fillets, the short side holds the 70 pt column with the card and tail.
One constant, not one per edge, because `fitZoom` divides by it to recover the
design size and cannot know the edge on its first frame. The extra area is
transparent and click-through; the hot-rectangle test is unchanged.
## Note on ordering

Cut from 3251208, before #223 landed. That one removed the stray `remoteSwitch`
fragment that stopped the settings script parsing, so the Edge and Screen
controls this adds are only reachable on top of it — they rebase cleanly.
Dragging and `config.json` drive the edge and the monitor either way.


vinzdg:
Merged in cb0ee53 (GitHub didn't pick it up because the README needed a hand merge with #229's tray docs). One follow-up on top: the notch window on the top and bottom edges was still 360 px wide, which clipped the pill once a fifth provider (Grok, #236) was on, so the flat window is now as wide as it is tall (a0b16b5). Thanks @1boooook2-maker!

PR #235, opened 2026-09-17
feat: support monitoring quota limits for multiple Antigravity accounts (#186)

### LINKED ISSUE #186: feat: Support monitoring quota limits for multiple Antigravity accounts
### Problem or Motivation

Hi!

Codenotch already supports separate Claude Code and Codex profiles. Would you consider adding support for monitoring multiple Google Antigravity accounts simultaneously?

I am not asking for account switching, automatic rotation, or running multiple Antigravity sessions. I only need separate read-only quota rings for each account, for example:

- Antigravity Personal
- Antigravity Work
- Antigravity Account 3

Each account could have its own credentials/profile path and display its own quota and reset time.

Is this something you plan to support, or would it be technically feasible in the current architecture?

I would be happy to test the feature.

### Proposed Solution

Add an Antigravity profile abstraction similar to the existing ClaudeProfile and CodexProfile implementations.

Each profile should have:

- a unique provider ID;
- account name/email;
- its own credential source;
- its own quota polling state and cached readings.

Codenotch could create one AntigravityProvider instance per discovered profile and poll the existing Google quota endpoint independently for each credential. The accounts could then appear as separate rings in Settings and on the notch, for example “Antigravity Personal” and “Antigravity Work”.

This feature does not need account switching, automatic rotation, or multiple Antigravity sessions. It only needs read-only quota monitoring for several accounts. The local language-server bridge could remain an optional source for the currently active account, while direct quota requests would be used for additional profiles.

### Alternatives Considered

_No response_

### Additional Context / Mockups

_No response_

divingsbysangam:
## Summary
Resolves #186 by introducing multi-account profile discovery and quota monitoring for Google Antigravity, following the architecture and design patterns established by `ClaudeProfile` and `CodexProfile`.

Users can now monitor multiple Antigravity accounts simultaneously (e.g. personal and work) with independent quota rings and activity tracking.

## Key Changes

1. **`AntigravityProfile` Discovery & Identity**:
   - Discovers configuration directories matching `~/.gemini/antigravity-<slug>` alongside the default `~/.gemini/antigravity`.
   - Filters out internal tool flavours (`ide`, `cli`, `backup`) from being recognized as separate accounts.
   - Preserves default provider ID `gemini` and display name `Antigravity` for backward compatibility with archives and user preferences.
   - Additional profiles receive provider ID `gemini-<slug>` and display name `Antigravity (<slug>)`.
   - Provides helper methods `isAntigravity(providerID:)` and `slug(fromProviderID:)`.

2. **Isolated Credentials & Thread-Safe Caching**:
   - Updated `AntigravityCredentials` with a thread-safe `profileCaches` map to prevent token and expiry cross-talk between accounts.
   - Supports reading credentials per profile directory from `profile.authURL` (`oauth_creds.json`), the login keychain (`profile.keychainAccount`), and `profile.configDirectory/agent.db`.

3. **Quota Provider & Activity Monitor Isolation**:
   - Updated `AntigravityProvider` to accept `profile: AntigravityProfile = .default()`.
   - The local language-server bridge (`localQuota()`) remains exclusive to the default profile, while non-default profiles poll Cloud Code PA directly using their respective credentials.
   - Updated `AntigravityActivityMonitor` with a convenience initializer `init(profile:)`, scoping transcript roots to `[profile.brainDirectory]` for named profiles while the default profile continues to monitor all install flavours.

4. **Integration**:
   - `AppDelegate`: Discovers all profiles at launch, registers `AntigravityProvider(profile:)` in `allProviders`, and sets up per-profile activity monitors.
   - `Preferences`: Added Antigravity profiles to `isDefaultOnFamily`.
   - `ProviderAccount`: Added Antigravity profiles to `usesKeychain`.
   - `UsageModel`: Added custom `authPrompt` guidance for `gemini-<slug>`.
   - `SettingsView`: Enabled notch reads and model limit dropdowns for all Antigravity profiles.

5. **Test Coverage**:
   - Added `AntigravityProfileTests` covering default identity, named profiles, slug extraction, ignored flavours, profile discovery, credential filter, and activity monitor roots.
   - All 8 new unit tests pass.
   - All 36 Antigravity-related tests pass.
   - All 1543 tests in the full test suite pass with 0 failures.

Closes #186


vinzdg:
Thanks @divingsbysangam — multiple Antigravity accounts is a good idea, but one thing blocks it: `AntigravityProfile.isAntigravity` matches the prefix `gemini-`, which also matches the existing `gemini-api` provider. That would make `gemini-api` default on for new installs (and switched on by `reconcile`), flag it as keychain-backed, show Antigravity's limit picker in Settings for it, and give it the text "Sign in to Antigravity in ~/.gemini/antigravity-api". A folder named `~/.gemini/antigravity-api` would also replace the Gemini API activity monitor. Please use a prefix nothing else uses, e.g. `antigravity-<slug>`.

Also:
- Adding Antigravity to `isDefaultOnFamily` makes it on by default for new installs; only Claude and Codex default on.
- Could you point to where Antigravity itself stores extra accounts as `~/.gemini/antigravity-<slug>` folders and keychain items under account `antigravity-<slug>`? I couldn't find it, and I'd rather not guess at another app's storage.
- `forgetCachedCredential` only calls `askAgain()` for the default account, so "Allow access…" can't bring the keychain prompt back for an extra one.
- `discover()` counts any folder with `settings.json` or `state` as an account; please require credentials.
- The new "Sign in to Antigravity in ~/.gemini/antigravity-%@…" string needs a catalog entry, and `signInRoute`/`sourceName` text should go through `L10n.t`.
- A test for the `gemini-api` clash and the default-off rule would lock this down.

The full suite passes on your branch, so it's close.

divingsbysangam:
Thanks for the thorough review @vinzdg! I've updated the PR to address each point:

1. **`gemini-api` Clash Prevention**:
   - Changed extra profile IDs from `gemini-<slug>` to `antigravity-<slug>` (e.g. `antigravity-work`), preserving `"gemini"` for the default account.
   - `AntigravityProfile.isAntigravity` and `AntigravityProfile.slug` now exclusively check for `"gemini"` and `"antigravity-<slug>"`. `"gemini-api"` returns false/nil and will never collide with Antigravity profiles.
   - Added `"api"` to `ignoredSlugs`.

2. **Default-Off for Antigravity**:
   - Removed Antigravity from `Preferences.isDefaultOnFamily`. Antigravity now starts disabled by default for fresh installs, keeping only Claude and Codex default-on.

3. **Account Storage & Credential Requirement in Discovery**:
   - Antigravity itself defaults to a single active configuration at `~/.gemini/antigravity` (`oauth_creds.json` / keychain `antigravity`). Users wanting to isolate multiple accounts follow the directory convention used by Claude (`~/.claude-<slug>`) and Codex (`~/.codex-<slug>`), placing their account's `oauth_creds.json` in `~/.gemini/antigravity-<slug>` or filing credentials under keychain item `antigravity-<slug>`.
   - To eliminate guessing, `discover()` now strictly requires valid credentials via `AntigravityCredentials.hasCredential(for:)` (checking `oauth_creds.json`, keychain item, or `agent.db` credentials). We also removed `"settings.json"` and `"state"` from candidate markers, so directories without genuine credentials are never picked up.

4. **"Allow access…" on Extra Accounts**:
   - Removed the `slug == nil` guard in `AntigravityProvider.forgetCachedCredential()`, so `AntigravityCredentials.askAgain()` is called unconditionally and re-enables interactive prompt permission for any profile.

5. **Localization**:
   - Added catalog entries for `"Sign in to Antigravity in ~/.gemini/antigravity-%@ to read your usage"` and `"Antigravity in %@"` with translations in `Localizable.xcstrings`.
   - Routed `sourceName` and `signInRoute` through `L10n.t`.

6. **Unit Tests**:
   - Added `testGeminiAPIClashPrevention`, `testAntigravityDefaultsOff`, `testDiscoveryRequiresCredentials`, and `testForgetCachedCredentialGrantsPromptPermission`.
   - All tests pass with zero failures.

vinzdg:
Thanks @divingsbysangam — the blockers are fixed: the `antigravity-` prefix no longer clashes with `gemini-api`, Antigravity stays off by default, discovery needs a credential, and "Allow access…" works per profile. Remaining before merge:
- The two new catalog keys are missing **uk** and **zh-Hant** (both otherwise complete).
- The one `PromptPermission` is shared by every profile, so another profile's background read can spend an "Allow access…" grant meant for this one — one per profile, please.
- The "Notch reads" picker now appears on every Antigravity profile but changes one global setting.
- Could you point to where Antigravity itself stores extra accounts' keychain items as account `antigravity-<slug>`? I'd rather not rely on a naming scheme the app never writes.
It will also need a small rebase against #272's AppDelegate change if that lands.

divingsbysangam:
Thanks for the review! I've addressed all feedback in commit `ea735be`:

1. **Missing localizations**: Added `uk` and `zh-Hant` translations for `"Sign in to Antigravity in ~/.gemini/antigravity-%@ to read your usage"` and `"Antigravity in %@"`.
2. **`PromptPermission` isolation**: Replaced the static shared `PromptPermission` in `AntigravityCredentials` with a thread-safe registry keyed by profile ID, and updated `AntigravityProvider.forgetCachedCredential()` to pass its specific profile so background polls cannot spend another profile's grant. Added a unit test covering this isolation.
3. **"Notch reads" / "Model data" pickers**: Reverted the check in `SettingsView` to `provider.id == AntigravityProfile.defaultID` so the pickers only appear on the primary Antigravity row and do not conflict over global preferences.
4. **Speculative Keychain naming**: Removed Keychain lookups for secondary profiles—extra profiles now strictly rely on `oauth_creds.json` and `agent.db`. Updated `AntigravityProfileTests` to assert `nil` keychain attributes for named profiles.
5. **Rebase note**: Keeping an eye on #272 to rebase if/when that lands.

vinzdg:
Merged as `7fceaf7`, closing #186. Thanks @divingsbysangam — you addressed everything from the last round properly: the `antigravity-` prefix no longer clashes with `gemini-api`, it stays off by default, and both new keys carry all eight locales including `uk` and `zh-Hant`. I re-ran the format-specifier check across the catalog and it's clean, with insertion order preserved. `PromptPermission` being a per-profile keyed registry (`AntigravityCredentials.swift:36-60`, `askAgain(for:)` at `:74`) is the right shape.

One follow-up of mine on top, `747e56b`: `ProviderSummary.usesKeychain` claimed every `antigravity-*` id was keychain-backed, but `read(for:)` only reads the profile's own `oauth_creds.json` / `agent.db` — the keychain is only touched by the default profile's `read()`. So extra profiles were being offered an "Allow access…" route back from a prompt that can never appear. Narrowed it to `AntigravityProfile.defaultID`.

One open question I'd still like answered, and it's the reason I hesitated: **is `~/.gemini/antigravity-<slug>` a layout Antigravity actually creates?** Last round the answer to the keychain question was to remove the keychain lookups rather than show that the convention exists. If those directories only appear when someone builds them by hand, this works but nobody will ever trigger it. If you can point at Antigravity writing them — a doc, a release note, a screenshot of a real config dir — I'll add it to the README so people know the feature is there. If it turns out it doesn't, tell me and we'll reshape it into something that keys off whatever Antigravity really does.

Nit for later: `AntigravityProfile.discover()` runs at `AppDelegate` init and opens each candidate's `agent.db` through SQLite on the main thread at launch.

PR #239, opened 2026-09-17
Windows: let the notch hide until the cursor reaches its edge

1boooook2-maker:
Settings' **Show** choice becomes three: **Always show**, **Auto-hide**, **Hide**.
Auto-hide keeps the pill off the screen until the pointer reaches the edge it is
pinned to, then slides it back out — the way the taskbar does. Hide stays exactly
what it was: the notch off altogether, tray icon only.

The pointer watchdog already samples the system cursor every 50 ms and already
holds the window's corner and the pill's own rectangle, so the reveal costs no
second poll and no new thread.

## Three choices worth stating

- **The strip that summons it is the pinned edge only, and only along the pill's
  own extent.** A band down the whole screen edge would pop the notch out every
  time somebody reached for a scrollbar. Move the pill to another edge and the
  strip moves with it.
- **It waits 900 ms before sliding away**, three times the card's own grace
  period. Losing the pill the moment the pointer slips off it, while reading the
  card, is what makes an auto-hiding strip tiring to live with.
- **Auto-hide is not "off"**, so it does not hold the tray icon on the way Hide
  does — the notch is still reachable at its edge.

Before the page has reported a rectangle — an auto-hidden notch that has never
rendered — the whole window edge answers instead, so it cannot end up
unsummonable.

## Update: a way in from the tray

The reveal strip is the notch's own edge on the notch's own screen — the right
rule while you know where it is, and no help at all when you do not. With three
monitors there is no sweeping every edge of every screen to find it.

So the tray menu offers it: one item, the notch comes out for five seconds, the
same idea (and the same default) as the Mac's `peek`. A menu item rather than a
click on the icon — the left click already opens the menu, and taking it for a
peek would spend the gesture that now shows the readings.

The watchdog holds the notch open for the length of the offer and then hands it
back to the hover fold; move onto it meanwhile and the normal grace period takes
over, so a peek that turns into a read does not snatch itself away. The item only
appears while the notch hides itself, and the menu's dedupe key carries that flag
— keyed on the readings alone it kept the stale menu.
## Note on scope

The Mac app has `peek`, which offers the notch for five seconds when a session
wants attention; it has no auto-hide. So this is a Windows-only setting rather
than a port, and it is off by default — nobody's notch changes behaviour until
they pick it. If you would rather the Windows port not grow a setting the Mac
does not have, say so and I will close it; no hard feelings.

## Verification

Right edge, five providers: hidden at startup with the cursor away
(`autohide: hidden again`), revealed on touching the right screen edge level with
the pill (`autohide: revealed at the edge`), hover card opening normally once out.
Four unit tests cover the strip — beside the pill only, at the edge only, per
edge, and the never-rendered fallback. `cargo test` on `windows/`: 92 passed,
0 failed.


vinzdg:
Thanks @1boooook2-maker — auto-hide is a nice idea, and it's off by default with a backward-compatible config, which is right. A few things before it can go in, and main has moved (#244 and #246 just landed), so it needs a rebase:

1. **A hidden notch never tells you when a session needs you.** Hook events still reach the page, but nothing calls `peek_notch` on attention. That's exactly what the Mac's peek is for; please peek when a session goes to attention.
2. **It fights the new handle carry (#246).** The auto-hide check in `main.rs` ignores `DRAGGING`, and carrying the notch moves the pointer off it, so the notch hides about 900 ms into a carry. Please skip auto-hide while dragging or carrying.
3. **It folds under the right-click menu (#244)**, since the pointer is on the menu rather than the notch.
4. **Settings text:** "Auto-hide" and its explanation are only in the Korean table; please add ru, zh, ja and uk.

When rebasing: keep #244's `refresh_provider`/`refresh_all` in tray.rs and add your `peek` item alongside the menu-text comparison, and keep both sets of new keys in i18n.rs.

RawJat:
Tried this build on Windows 11 (the PR head as written, on its 1.13.0 base). The 900 ms grace period is the right call — losing the pill while reading the card would have made it unusable — and limiting the reveal to the pill's own extent rather than the whole edge is the detail most auto-hides get wrong.

Two things I didn't see in the list above:

**1. Auto-hide plus the tray icon off leaves nothing on screen.** `set_ui_flags` forces the tray on only when `notch_visible` is false, so `notch_visible = true, notch_autohide = true, tray_visible = false` is reachable from Settings and leaves no notch, no tray icon, and an invisible strip as the only way back. The follow-up commit's tray item can't help, because there is no tray icon. Simplest fix is to treat auto-hide like "off" for that one guard: `c.tray_visible = if notch_visible && !notch_autohide { tray_visible } else { true }`.

**2. `REVEAL_BAND = 4.0` is a very small invisible target.** For comparison, the Mac's folded notch keeps a *visible* resting pill (`NotchLayout.pillWidth` × `pillHeight`, ~10 × 79 pt, commented as "sized to read as a deliberate handle rather than a sliver of chrome") and gives it a wake zone *larger* than itself (`pillHotZone`, ~34 pt). Widening the band would help on its own; a visible marker is a bigger change and probably belongs with a fold rather than here.

On the two collisions with our merged PRs, since it is our code:

- **The carry (#246):** `begin_move` and `drag_begin` share the same `DRAGGING` static (`main.rs:313`), so one guard in the auto-hide block covers both — `if DRAGGING.load(Ordering::SeqCst) { away = 0; continue; }`.
- **The right-click menu (#244):** `notchmenu.rs` has no open-state flag, so this one needs a new one set around `show_notch_menu` and cleared when it closes. While the popup is up the pointer is genuinely off the notch, so nothing existing can tell the difference.

Separately, and so nobody writes the same watchdog twice: I'm looking at porting the Mac's fold (the visible resting pill) as its own change. It is a different behaviour from hiding, not a replacement for this.

vinzdg:
Closing this one, and I want to be clear it's not because it was the weaker patch — #288 has now landed and covers the same ground.

For what it's worth, when I compared the two auto-hide PRs directly, yours was the better of them: the reveal strip computed per edge on the Rust side, the tray peek, off by default, and four real unit tests. #283's only advantage over it was merging cleanly. Both lost to #288 for a different reason — it puts the behaviour as a third mode on the existing **Show** row rather than adding a second control beside it, which is the shape the Mac has, and it mirrors the Mac's `foldGrace` and `pillHotZone` values directly.

@RawJat's note on your build was right too — the 900 ms grace period was the correct call, and #288 ended up at the same place.

Thanks @1boooook2-maker. If you want to pick up the tray peek separately, that part has no equivalent in what landed and I'd take it.

PR #248, opened 2026-09-18
Add a QianwenAI ring, read from the platform console's own gateway

LordCasser:
Adds a QianwenAI provider. The ring shows the personal Token Plan's 7-day credits window, read from the platform console after signing in inside Codenotch.

### Why it reads the console

The platform publishes model-call APIs and nothing for usage, so the Token Plan's own keys cannot say how much of the allowance is left, and the docs allow those keys only for interactive use in coding tools. The console is where the number lives, and its analytics page reads one RPC gateway. This follows DeepSeek, MiniMax and Perplexity: a WKWebView session the user signs into themselves, no browser cookies read, and no request at all before the first sign-in.

### The request

```
POST https://cs-data.qianwenai.com/data/api.json
product=sfm_bailian&action=BroadScopeAspnGateway&sec_token=…&region=cn-beijing
&params={"Api":"zeldaHttp.apikeyMgr./tokenplan/personal/api/v2/usage","Data":{},"V":"1.0"}
```

`sec_token` comes from `GET https://platform-home.qianwenai.com/tool/user/info.json`, read fresh every time so a token that has aged out cannot pass for an empty reading. The `Api` inside the JSON `params` is what routes the call, which the query string does not make obvious: dropping that field made the gateway stop echoing it back.

The ring reads `per1WeekPercentage` and `per1WeekResetTime`, the two fields the console's own individual-plan card draws. Nothing is derived from them, and a payload carrying neither that fraction nor credit counts reports `nothingMetered` rather than inventing a denominator.

Signed out, the gateway answers HTTP 200 with `{"code":"ConsoleNeedLogin"}`, so the body decides and never the status. Only that marker becomes 401. A 500 or a malformed envelope keeps its status and reaches the parser, which renders an error and keeps the last reading; 401 discards the remembered reading and would send someone who is signed in to go and sign in over a server fault.

### Sign-in

The platform signs in through a redirect chain that leaves the console origin: `account.qianwenai.com`, then `account.aliyun.com`, then back to `/home/`. The sheet used to probe its session only when the window closed, so closing it while one of those pages was still on screen left someone signed in with the app still asking. It now probes while the sheet is open, once per page rather than once per tick, because a sign-in completes by navigating back and a blind tick put forty probes a minute on a window that can sit open while someone fetches a verification code. That change is in `WebSessionProvider`, so DeepSeek and MiniMax inherit it.

### Icon

The glyph is the console's own favicon. That icon is a white mark on a blue gradient plate, and `ProviderGlyphView` tints what it draws, so plate and mark would come out one colour and render as a solid rounded square. The plate, its gradient and the clip are dropped and the viewBox is cropped to the mark, as Devin's asset did with its plate. Measured off a 512px raster, the mark's ink fills 0.996 of its box, so `opticalScale` is 0.97, which brings it to Claude's extent. It is a different emblem from the Lobe Icons `.qwen` that local model cells draw, so a platform ring cannot be mistaken for a model cell.

### Verification

`make test` was not run. The machine this was written on has no Xcode and no xcodegen, so the suite cannot run there at all. CI is where it will actually run.

What was run instead:

- `swiftc -typecheck` over the touched files against the real sources, including `Sites.swift`, `QianwenUsage.swift` and `WebSessionProvider.swift` with their dependencies. Clean, and no new warnings against `HEAD`.
- The parser compiled and executed against every case the tests assert, plus the failure envelopes.
- The site script executed in a stubbed browser: signed in, signed out, a failed token call, a `ConsoleNeedLogin` body under HTTP 200, and a server-side failure, checking the form fields, the routing field and the status mapping each time.
- The gate the new poll drives, running the real `WebSessionAuthenticationGate`.

None of that covers `make test` itself, the WKWebView and CORS behaviour against the live console, the SSO return, or the shape of a real payload.

### Still open

No authenticated response has been captured from the platform yet. The test fixtures were read out of the console's shipped JavaScript and the numbers in them are invented, so they pin the shape this code was written against rather than the platform's real one. The first sign-in logs the body verbatim (`qianwenai usage -> …` in the unified log) and those fixtures should be replaced with it. That is also where a 5-hour window would show up, if the personal plan has one.

Left out on purpose: the international site (`qwencloud.com`, the same gateway under `IntlBroadScopeAspnGateway`), add-on credit packs and reset cards, the usage trend, and the Windows port. The provider starts switched off in Settings, as every provider except Claude and Codex does; switching it on opens the sign-in sheet.


LordCasser:
Heads up on the checks: both workflows are sitting in action_required, which is the first-time-contributor gate on fork pull requests, so nothing has run yet. The suite has not been run locally either, for the reason in the description, so it needs an Approve and run to actually run.

vinzdg:
Merged, thank you @LordCasser — nicely done, and the envelope handling and tests are solid. I checked `qianwenai.com` is Alibaba's (their registrar and DNS) before sending a session there. Follow-ups in 1a8f75a:
- **Sign-in polling is now opt-in per site.** The PR made every site with a probe poll while its sign-in window is open, which brought back what #172 removed for DeepSeek and MiniMax (real API calls every 1.5 s to endpoints that have throttled the app). QianwenAI keeps it; a test holds the others to confirm-on-close.
- Responses aren't logged verbatim for QianwenAI, since the envelope's other fields are undocumented.
- Sign-out also clears the Aliyun SSO cookie, so the next sign-in can't silently reuse the old account.
- The ring reuses the translated "Weekly limit" label, and the two messages are translated.

One thing worth confirming against a live account: `per1WeekPercentage` is treated as a 0–1 fraction. If the gateway ever sends 42 rather than 0.42, the ring would read 100%.

PR #252, opened 2026-09-18
Translate the release notes into Japanese

urib0:
The release-note history had no catalog keys at all — 103 of them,
covering 1.7.0 through 1.13.1. Five more keys existed but had no
Japanese: the Usage Limits section added in 1.12.0.

With these, every literal `L10n.t` key in `Sources` has a catalog entry
and a Japanese value: **704 of 704**.

### Translation notes

`Watch limit` and `Critical limit` are rendered しきい値 (threshold)
rather than 上限 (limit). They name the points where a ring turns amber
and red, not a quota — and Ukrainian, the only other complete locale,
reads them the same way (поріг).

Product and language names stay exactly as written: Deutsch, Русский,
简体中文, Українська, Kimi, LM Studio, Ghostty, Windows. "日本語 and
Português (Brasil)" keeps both names and translates only the
conjunction.

### What this does not touch

Only `ja` is added or filled in. No `fr`, `pt-BR`, `de`, `ru`, `uk` or
`zh-Hans` value is modified — verified at the JSON level, along with the
key set and its order. The diff is 1163 lines, all additions.

### Checks

Every key was checked mechanically against its English source for
matching argument order, `%%` count and bare `%` count.

`make test` passes: 1577 tests, 0 failures, 3 skipped.

### Thanks

Two things from earlier rounds landed well from this side: the crash in
the Simplified Chinese threshold string is fixed with positional
specifiers in 1.13.1, and the note says every translation is now checked
for it. That check is the part that matters — it is the class of bug, not
the one string.

https://claude.ai/code/session_0186RtTxiXLXiuZTxwZ3sDD9


vinzdg:
Thanks @urib0 — the Japanese release notes look good and every key matches `ReleaseNotes.swift`. The branch predates 1.14.0, so the 11 strings in the 1.14.0 notes ("A new Settings", QianwenAI, 繁體中文, the Windows notch, Fresh logos, the headline and their details) are missing. Could you rebase and add those? Then it's ready.

vinzdg:
Merged as `ebb0201`. Thanks @urib0 — I checked this carefully because the catalog is the one file that has crashed a release: 103 new keys plus 5 `ja` values on existing ones, insertion order preserved, zero real deletions, entry shape matching the rest. Every added string is argument-free, so there's no specifier ordering to get wrong, and the two `ja` values carrying a literal `%` mirror English keys that do the same with no arguments, so `CatalogFormatTests.formattedKeys` correctly ignores them.

Landing it as-is rather than holding it: 1.0.0 through 1.13.1 are fully covered, and the 21 strings from 1.14.0 and 1.15.0 fall back to English harmlessly in the meantime. Opened an issue to track those — no obligation to pick it up yourself, though you'd obviously be the best person to.

This is the kind of contribution that's easy to undervalue because it doesn't change behaviour. It matters; thank you.

vinzdg:
Tracking issue for the remaining 21 strings: #290

PR #255, opened 2026-09-18
Add runtime provider plugins (manifest + exec protocol)

8nevil8:
## Summary

External tools can now register usage providers at runtime, without being compiled into Codenotch. A plugin is a `plugin.json` manifest plus an executable; Codenotch polls it and renders its JSON snapshot natively — ring, tooltip, settings row, ordering, connect/disconnect and notifications all behave exactly as for built-in providers.

Plugins live with their vendors, not in this repo. The reference implementation ships as codemie-ai/codemie-code#569: `codemie install codenotch --budget-plugin` installs this app and registers two CodeMie providers (Budget with a Total headline; Claude with per-bucket session spending + live activity) from the user's existing CodeMie CLI setup — no separate sign-in, no separate binary.

## Changes

**Plugin runtime** (`Sources/Plugins/`, wiring in `AppDelegate`)
- `PluginManifest` — schema v1 validation (id shape, built-in collision, absolute+executable exec, glyph presence)
- `PluginRegistry` — scan + validate + `DispatchSource` watch; live register/deregister; `CODENOTCH_PLUGINS_DIR` dev override
- `ExternalPluginProvider` — actor `UsageProvider` over an exec JSON bridge (same pattern as `ClaudeUsageCLI`); protocol exit codes → `UsageProviderError`; account from payload
- `UsageStore.register/deregister` — dynamic providers with archive pruning; never-seen plugin ids start connected (registration is explicit), toggle-off persists
- `ProviderGlyph.external` + `PluginGlyphStore` — archive-safe plugin marks (closed Codable enum untouched by per-plugin cases)
- `ActivityCoordinator.setMonitor/removeMonitor` — plugin manifests can attach a `claudeSessions` monitor, so wrapped agents get live session rings
- Contract: `docs/design/plugin-protocol.md`; design + alternatives (HTTP push, in-process bundles) + trust model: `docs/specs/2026-09-18-provider-plugins-design.md`

## Testing

- 45 new app tests (manifest, payload mapping, exit codes, real spawn + timeout kill, registry watch, store register/deregister). These five files use **swift-testing** rather than XCTest: CLT-only machines ship no `XCTest.framework`, and swift-testing runs both there and inside the Xcode test target. Rest of the suite untouched.
- `make build` + `make test-ci` (full suite) green on Xcode 27.
- Verified end-to-end with the vendor plugin (codemie-ai/codemie-code#569) on a real CodeMie SSO account: both rings in the notch with real values, settings rows with account identity, disable/enable, and unregister→register — all without restarting the app.

## Checklist
- [x] Self-reviewed
- [x] Manual testing performed (live app on real CodeMie SSO account)
- [x] Documentation updated (protocol doc, design spec, README, TASKS)
- [x] No breaking changes (additive; built-ins untouched)


vinzdg:
Thanks @8nevil8 — this is a thoughtful design and well documented, but I can't accept it in this form: it turns a notarized app into a launcher for any program described by a JSON file in a folder, switched on automatically.
- Any process running as the user can drop a manifest into `~/Library/Application Support/Codenotch/Plugins` (or `CODENOTCH_PLUGINS_DIR`), and it is picked up within half a second and **enabled without asking**, then run every refresh under Codenotch's identity (so macOS attributes its prompts and grants to Codenotch).
- `exec.path` can be anything, including `/bin/sh -c …`; `signIn.run` isn't validated at all.
- The child inherits the app's whole environment (e.g. `GITHUB_TOKEN`).
- The timeout has no ceiling and only sends SIGTERM; output is read without a size cap.
- `displayName` is unchecked, so a plugin can present itself as "Claude"; `glyph.image` can escape the plugin folder with `../`; `manageURL` accepts non-https schemes that `NSWorkspace.open` will follow.

If you'd like to take it further, the minimum I'd need: off by default with explicit per-plugin approval in Settings showing the exec path and a hash (re-asked on change); refuse group/world-writable or non-owned folders and symlinks, and ignore the env override in release builds; a minimal environment; a capped timeout with SIGKILL; capped output and validated fields; plugin rows visibly labelled as plugins; https-only links and glyphs kept inside the plugin folder. I'm closing this for now — happy to discuss a smaller first step.

PR #258, opened 2026-09-18
Improve Liquid Glass contrast

liuzel01:
Improve Liquid Glass contrast

## Summary

Improve readability and visual consistency for Liquid Glass UI components.

The change:

- Adjusts the Liquid Glass palette for clearer contrast
- Improves tooltip and usage detail presentation across appearance modes
- Adds usage-band test coverage

## Testing

- [x] 1611 tests passed
- [x] 7 tests skipped
- [x] 0 failures

## Related commit

`7dbca5e`


vinzdg:
Thanks @liuzel01 — the dim under dark-mode glass tooltips is a good idea. One problem: the new `Palette.tooltipTextSecondary` (#C2C2C2 in dark) replaces the secondary ink everywhere, so Solid and Dark glass tooltips lose their frame-accurate #808080 too, contradicting the PR's own comment. Could you choose the brighter ink only when `TooltipGlassContrast.needsReadableDim` is true (e.g. through an environment value)? Also note that standard glass in dark mode now gets the same 0.60 dim as Dark glass, so the two styles look nearly alike in tooltips — worth a sentence in the PR on whether that's intended.

liuzel01:
Updated the existing PR after rebasing onto the latest main.

- Restored frame-accurate #808080 secondary ink for Solid and Dark glass.
- Limited brighter secondary ink to dark standard Liquid Glass only.
- Kept Dark glass readable by deepening the tooltip-only backing to 0.80, while its main notch backing remains 0.60.
- Manually tested the updated Dark glass tooltip contrast.

Could you take look.

vinzdg:
Merged as `5ac523a`. Thanks for turning this around so quickly @liuzel01 — the gate is exactly what I wanted: `TooltipGlassContrast.needsReadableDim` keys on `effective == .glass && colorScheme == .dark && !reduceTransparency`, every call site is already inside an `if glassy` branch, and `.solid`/`.darkGlass` keep their frame-accurate #808080. On macOS 15 `effective` is always `.solid`, so the opaque appearance is untouched. The test matrix in `UsageBandTests.swift:67-89` pins all of it.

I looked hard at the `.darkGlass` tooltip dim going 0.60 → 0.80, since that's a visible change for existing dark-glass users. Keeping it: it's tooltip-only (the notch surface still reads `glassDim` at `NotchRootView.swift:175/185`), and the deeper backing is there precisely so the #808080 ink you preserved stays legible against a light desktop through `Glass.clear`. The reasoning in the doc comment holds up.

One small follow-up of mine, `de5fbde`: the new dim assertions sat behind `guard let … else { return }`, so they'd have passed vacuously if `resolve` returned nil. Switched to `XCTUnwrap`.

Left alone: `UsageResetCard.swift:18-21` recomputes the ink locally rather than reading `\.tooltipSecondaryInk`, duplicating the rule `NotchRootView.swift:136` publishes. Worth collapsing next time you're in there.

I've also left notes on #273 — same idea, needs one more pass.

PR #261, opened 2026-09-18
Make Accounts sidebar row expandable

liuzel01:
## Summary

Improve the Accounts section in Settings by allowing the sidebar row to expand and collapse its provider entries.

The change:

- Adds an expandable Accounts sidebar row
- Shows or hides configured provider entries from the row interaction
- Preserves the existing provider navigation behavior

## Testing

- [x] 1608 tests passed
- [x] 3 tests skipped
- [x] 0 failures

## Related commit

`2c9e9d7`


vinzdg:
Thanks @liuzel01. Since 1.14.0 the sidebar was redesigned, and the Accounts arrow is deliberately its own control (hover disc, press, spring turn). This PR turns it into a plain indicator and makes the whole row toggle — including when you arrive at Accounts from another pane, which folds the providers away every time. I'm going to close this one; if the arrow's target feels too small, a smaller change (toggle only when Accounts is already selected) would be welcome as a new PR.

PR #272, opened 2026-09-19
Draw one Claude ring per organization, not per directory

mantovaniai:
## The problem

Claude profile discovery is keyed on the configuration directory (`~/.claude` and each `~/.claude-<slug>`), but a limit belongs to an **organization**. One login can front several: a personal Max plan and a Team one, switched between in Claude Desktop's account menu, each with its own limits. `ClaudeProfile` already knows this, and `organizationName()` says as much in its doc comment.

On the machine this was found on, `~/.claude` and `~/.claude-vcore` were both signed into the same Team organization. The two rings carried byte-identical windows, down to the same `resetsAt` to the fractional second, because both asked the Desktop cache for the same entry. Meanwhile the personal organization, sitting at **98% of its session**, had no directory of its own and so no ring at all.

That is the worst shape a blind spot can take: two Claude rings read as two accounts covered.

## The change

`ClaudeProfile.discover` now keeps one profile per organization.

* Only a **readable** uuid merges. An organization that cannot be read is not evidence of anything, and treating `nil` as a value would collapse every unreadable profile into a single ring.
* Which profile survives is decided twice. A profile that still has a token outranks one that does not, because signing out leaves `.claude.json` behind with the organization still in it, and preferring that stale default would merge a working login into a ring that can never refresh. Among equals the input order stands (the default first, then slugs alphabetically), so the surviving ring keeps the id its archived readings and connection choices are filed under.
* `hasCredential` is asked only on a collision, and it enumerates keychain attributes rather than reading a secret, so it costs no prompt.
* Each merge logs a notice naming both directories and the organization.

### Sessions

The merge cannot simply forget the directory it drops: Claude Code still runs there, and those sessions belong to the ring that absorbed it. Discovery gained `discoverGrouped`, which carries the merged directories through, and `AppDelegate` watches all of them under the one id.

That needed a combining monitor. It combines with `combineLatest` rather than `merge`, because `merge` would publish one directory's sessions as though they were the whole ring and blank out the other's, trading the duplicate ring for a quieter version of the same blind spot. A group of one, which is every install without a duplicate login, gets its `ClaudeSessionMonitor` directly, unwrapped, so the ordinary path is byte for byte what it was.

### Known limit

Discovery still runs once at launch, so signing a directory into a different organization takes a relaunch to become its own ring, exactly as adding the directory does. Noted in the `AppDelegate` comment rather than worked around.

## Testing

* `make test-ci` on top of current `main`: 1622 tests, 0 failures.
* 7 new tests, in `ClaudeProfileTests` and a new `CombinedClaudeSessionMonitorTests`.
* Written test first: the three behavioural tests failed before the change, and the rest passed throughout as regression guards.
* Verified on a real install. Before: two rings both reading 4% / 1%. After: `Claude Team` at 14% / 3% and `Claude Personal` at 98% / 12%, the second of which had no ring at all before.

Test fixtures use synthetic organization uuids.


vinzdg:
Thanks @mantovaniai — the problem is real (one org showing twice), and the session merging is nicely done. Two things block it:

1. **The merge key is too broad.** Grouping by `organizationUuid` alone merges two different people in the same Team or Enterprise org into one ring, and each of them has their own limits, so one person's usage disappears. The key needs to be account + org (e.g. `accountUuid` or email together with `organizationUuid`); `AccountFile` doesn't decode `accountUuid` yet.
2. **Ring ids can change under the user.** The surviving directory is chosen by `hasCredential`, so signing out of one and relaunching can flip the ring between `claude` and `claude-work`, taking its archived readings and alert mutes with it. And if someone has `claude` off and `claude-work` on, `claude` can win and the only ring they enabled silently disappears. Please prefer a connected id when picking the survivor (or carry the on-list entry over), and keep the choice stable across launches.

Also worth knowing: if the survivor's keychain service has a recorded Deny, the ring stays empty even when the merged directory's login would read fine.

vinzdg:
@mantovaniai — a heads-up that this now conflicts in `Sources/App/AppDelegate.swift`; #235's Antigravity multi-account support landed there today.

The three blockers from my review are still open, and the one I'd most want your thinking on is the first: the merge key is `organizationUuid` alone, so two people in the same Team or Enterprise org on one Mac collapse into a single ring. `AccountFile` already decodes `emailAddress`; keying on that alongside the org would fix it. The other two are the survivor selection (a ring can currently vanish for someone whose only connected profile loses the `hasCredential` race) and the missing migration for archived readings and connection state.

Happy to look again whenever you push.

vinzdg:
@mantovaniai — closing this, and I owe you a clear reason rather than silence, since it's been open since the 19th and you did careful work.

Two things decided it.

**The unit behind a Claude ring is the account, not the organization.** #321 landed today taking that route, and the two models contradict each other: `oneRingPerOrganization` (`ClaudeProfile.swift:137`) drops directories that share an `organizationUuid`, but two people on one Team org — or one person on two seats — have *separate* limits. Merging them collapses two real readings into one and loses a limit, which is the opposite of what a usage app should do. The duplicate-ring case you found is genuinely real; it's the join that had to change.

**There's no migration.** Archived readings, connection choices and hover bands are all filed under the profile `id` (`ClaudeProfile.swift:122`). A merged-away directory's id simply disappears, so an existing user's second ring would vanish on upgrade with nothing to bring it back. On its own that would be fixable; combined with the first point it isn't worth fixing.

It also no longer merges — `AppDelegate.swift` has moved under it several times this week.

**The part worth keeping:** if the same *account* is reachable from two directories, that genuinely is one ring shown twice, and a de-dupe keyed on `accountUuid` would be welcome. `ClaudeProfile` now has `accountID()` from #321, so the join you'd need already exists. If you want to come back to it on that basis I'll review it properly and quickly.

`CombinedClaudeSessionMonitor` was nicely done, for what it's worth — `combineLatest` rather than `merge` was the right call and I said so at the time.

PR #273, opened 2026-09-19
Keep Accounts disclosure arrow independent

liuzel01:
## Summary

Refine the Accounts sidebar interaction based on maintainer feedback from the previous Accounts expansion PR.

The change:

- Allows the remaining Accounts row area to expand or collapse providers
- Toggles expansion from the row only when Accounts is already selected
- Avoids changing the disclosure state when navigating to Accounts from another settings pane

## Testing

- [x] 1640 tests passed
- [x] 7 tests skipped
- [x] 0 failures
- [x] Verified the chevron and Accounts row interactions manually

## Notes

This PR supersedes the earlier Accounts sidebar expansion approach 

Previous PR:
[261](https://github.com/vinzdg/codenotch/pull/261)

## Related commit

`609b915`


vinzdg:
Thanks @liuzel01 — separating the chevron from the row is the right instinct, and the sidebar rework left this ambiguous. Two things before I can take it.

**1. It shrinks the row's hit target.** The PR removes `.contentShape(Self.pill)` from the outer row (`SettingsView.swift:239`) and puts `.contentShape(Rectangle())` on an inner `HStack` (`:206-215`). But the outer view carries all the padding — leading 10/28, trailing 8, vertical 6 at `:208-210` — and also the selection/hover pill background. So the pill's whole padding band goes dead to both clicks and `.onHover` (`:244`): the row gets a smaller target than its visible pill, and the hover highlight stops lighting up near the edges. Keep `.contentShape(Self.pill)` on the outer view and move the padding inside the gesture-bearing stack instead, or give the chevron hit-test precedence with `.highPriorityGesture`.

**2. It doesn't quite do what the title says.** `activate()` at `:247-258` means that when Accounts is *already selected*, clicking anywhere on the row body toggles the disclosure. So there are two controls doing the same job with nothing telling you so, and the `matchedGeometryEffect` pill gives no feedback for it. If the rule is "chevron toggles disclosure, row body selects" — which is what I want — then `activate()` should just be `select()`.

Nit while you're in there: `.accessibilityElement(children: .combine)` at `:265` swallows the `DisclosureChevron` button, so VoiceOver has no way to expand Accounts without also selecting it, and there's no `.accessibilityValue` or expanded trait. That's pre-existing, but this PR is the one that makes the two actions genuinely distinct, so it's the natural place to fix it.

Thanks also for #258 — that one's clear and going in today.

vinzdg:
@liuzel01 — still happy to take this; the two items from my review are small. The row's hit target and hover band shrinking is the one that matters, since the padding lives on the outer view that lost its `.contentShape`.

No rush, and thanks again for #258, which shipped in 1.16.0 yesterday.

PR #274, opened 2026-09-19
Translate the Windows notch placement settings into Russian

unchase:
## Summary

A Russian Settings window shows the notch placement rows in English: Edge with its four options and their captions, Screen, Main, and Show move handle with its caption. These strings reached the Chinese, Korean and Ukrainian tables, but not `RU_STATIC`.

## Changes

- `windows/codenotch/ui/settings.html`: add the 15 missing strings to `RU_STATIC`, using the terms the Russian table already uses (панель, край, карточка). Nothing else changes.

After this, every string the page shows has a Russian translation. The only keys still in other tables and not in `RU_STATIC` are two older wordings of the drag hint, which the page no longer shows.

## Test Plan

- [x] `node --check` on the page script: passes.
- [x] Opened the page in a browser with a stubbed `window.__TAURI__` and `ru`: the Appearance pane is fully Russian, with no console errors.
- [x] Built on Windows 11 and ran it.
- [ ] The Screen row with more than one monitor: not seen, since this machine has one monitor.


vinzdg:
Merged as `38eba73`. Thanks @unchase.

Heads-up on one thing I had to fix in the merge: #276 landed just before this and reworded two of the English sources you translated. `'Which screen the notch lives on. Dragging it onto another screen sets this too.'` lost its second sentence, and the pill-drag hint was replaced wholesale by the Alt-drag one. Merged as written, those two Russian strings would have been dead keys rendering English. I re-pointed the first at the shortened source and dropped the second, since #276 already carries Russian for its replacement.

I also added a check across all six `*_STATIC` tables — `RU_STATIC` is now at 81 keys with nothing stale and nothing missing that the other locales have.

PR #275, opened 2026-09-19
Add runtime provider plugins, hardened per the #255 security review

8nevil8:
## Summary

Re-lands the runtime provider plugins from #255, reworked end-to-end around @vinzdg's security review there. Every concern from that review is addressed; the point-by-point mapping is below. The design spec (`docs/superpowers/specs/2026-09-19-plugin-hardening-design.md`) and implementation plan (`docs/superpowers/plans/2026-09-19-plugin-hardening.md`) are committed in the branch, and the protocol doc (`docs/design/plugin-protocol.md`) documents the new contract.

## The review's concerns, one by one

> Any process can drop a manifest into the plugins folder and it is picked up within half a second and enabled without asking

Registration is no longer execution. A new or changed plugin is **not run**: it appears in Settings → Accounts under "Plugins awaiting approval" showing the display name, full exec path + arguments, and a SHA-256 over the `plugin.json` bytes followed by the executable's bytes (`PluginRegistry.contentHash`). Enabling pins that exact build; **any change to either file re-pends the plugin** (the hash is part of the registry's `Equatable` diff, and the watcher now covers the plugin subdirectories). Approvals persist across reinstalls of identical bytes. The auto-connect path is deleted — approval is the only way a plugin gets connected, at launch and at runtime alike. Pending rows also offer to open the manifest and reveal the folder in Finder, so the decision can be made from the files themselves.

> `exec.path` can be anything, including `/bin/sh -c …`; `signIn.run` isn't validated at all

`signIn.run` is now validated exactly like `exec` (absolute, exists, executable, trusted). `/bin/sh` remains an allowed executable by design — what changed is that it can no longer run *silently*: it needs approval of a hash that covers the exact bytes, and the settings row shows the full command line including args.

> The child inherits the app's whole environment (e.g. `GITHUB_TOKEN`)

The child environment is now a fixed allowlist — `PATH` (system dirs + Homebrew), `HOME`, `USER`, `LOGNAME`, `LANG`, `TMPDIR` — on both the fetch and the `signIn.run` spawn. Nothing is inherited.

> The timeout has no ceiling and only sends SIGTERM; output is read without a size cap

`timeoutSeconds` is clamped to 30 s; expiry sends SIGTERM, then SIGKILL after a 2 s grace. stdout is capped at 1 MiB, stderr at 64 KiB; exceeding either SIGKILLs the child and fails the fetch, and the drain wait is bounded so a pipe-holding grandchild cannot hang a refresh.

> `displayName` is unchecked; `glyph.image` can escape with `../`; `manageURL` accepts non-https schemes

`displayName` must be 1–40 chars, no control characters, and must not case-insensitively match a built-in provider's name (compared after trimming). `glyph.image` must resolve inside the plugin directory after `..` and symlink resolution. `manageURL` is https-only; anything else is dropped before `NSWorkspace.open`.

> The minimum I'd need: off by default with approval in Settings; refuse group/world-writable or non-owned folders and symlinks; ignore the env override in release builds; minimal environment; capped timeout with SIGKILL; capped output and validated fields; plugin rows visibly labelled; https-only links; glyphs inside the folder

All present. Folder/exec trust (`PluginTrust`): plugins root, plugin directories and manifests must be user-owned, non-symlink, not group/world-writable; executables may additionally be root-owned (system binaries). `CODENOTCH_PLUGINS_DIR` is honored in debug builds only. All plugin rows carry a "Plugin" badge.

Detection boundary, stated plainly rather than overpromised: the watcher covers the plugins root and one subdirectory level (~0.5 s debounce), so manifest edits and in-tree binary swaps re-pend live; an exec binary outside the plugin tree overwritten in place is caught at the next launch scan, and approval always pins the bytes as they are at approval time.

## Test plan

- `make test`: 1714 passed, 0 failed (macOS arm64). New suites cover trust checks, validation, hashing, approval flow (pend → approve → re-pend), spawn limits, and URL filtering — mostly against the real filesystem and the real directory watcher.
- Manual end-to-end on a live build: dropped a manifest → pending row with path/hash, nothing running → Enable… → connected with "Plugin" badge → edited the manifest → pending again with a new hash.
- Companion change in the CodeMie CLI installer (writes the resolved real executable path, since Homebrew/npm shims are symlinks and now refused) — not part of this repo.

## Collaboration

Thanks again for the review on #255 — it was specific enough to implement against directly, and the design it forced is better than what was there. Happy to keep working through this together: if you'd rather land a smaller first step (e.g. approval gate alone, with the rest staged behind it), or want any of the defaults tuned (timeout ceiling, output caps, env allowlist), say the word and I'll adjust.


vinzdg:
Thanks @8nevil8 — this is a serious pass and most of the #255 list is genuinely fixed. The environment is a fixed allowlist applied to both the fetch and `signIn.run` (`ExternalPluginProvider.swift:41-50`, `:147`, `:180`), stdin is `/dev/null`, nothing goes on argv, the timeout is capped with a real SIGKILL, output is capped, the glyph is confined, `manageURL` is https-only, and the env override is debug-only. Credit where it's due.

Two things still block it:

**1. The approval gate is stored where the attacker can write.** `pluginApprovals` lives in the app's UserDefaults plist (`Preferences.swift:27-29,414,833-836`). That plist is writable by exactly the same "any process running as the user" that drops the manifest in the first place — so that process can compute the SHA-256 itself and pre-approve its own plugin. The opt-in gate can be bypassed end to end with no user interaction, which puts us back at #255(a) in practice even though the flow is correct. This wants to live somewhere the attacker can't write — the keychain, or at minimum an HMAC keyed outside defaults.

**2. The hash doesn't cover the code that actually runs.** `PluginRegistry.swift:69-77` hashes `plugin.json` plus the bytes of `exec.path`. For the interpreter case the PR specifically documents — `/bin/sh run.sh`, `/usr/bin/python3 x.py`, `/usr/bin/env …` — that pins the interpreter, not the script. `run.sh` can be rewritten at will, the diff sees no change, and it never re-pends. The PR body's "a hash that covers the exact bytes" isn't true for the case it's defending against.

Related: `PluginManifest.validateExecutable:159-175` only requires absolute + executable + `PluginTrust`, so the exec can live outside the plugin tree (`~/Downloads/foo` at 755 passes), and the watcher only covers the plugins root plus one level — so a post-approval in-place swap goes unnoticed until the next launch, and nothing re-verifies the hash before `spawnSync`.

Also worth doing before this lands:

- **No revocation.** Nothing ever clears `pluginApprovals`. Deleting a plugin deregisters the provider but keeps the approval, so re-dropping identical bytes silently auto-runs. There needs to be a way to forget an approval in Settings.
- **The badge is Settings-only.** `ProviderSnapshot` carries no `isPlugin`, so the notch ring and tooltip show the plugin's own image and name with no marker. A plugin shipping a Claude-lookalike glyph named "Claude Max" is indistinguishable there — that's #255(c) in the place it matters most. The name check at `PluginManifest.swift:106-116` is also a literal comparison, so Cyrillic "Сlaude" passes.
- **Payload isn't bounded.** `PluginSnapshotPayload` accepts unbounded `windows` count and string lengths, unclamped `usedFraction`/`money`, arbitrary `resetsAt` (which drives reset notifications) and unclamped `retryAfterSeconds`.
- The id namespace prefix from requirement 3 wasn't implemented, and the built-in collision set is frozen at launch.

On scale: ~1,100 source, ~1,460 test and ~2,280 doc lines. I'd drop `docs/superpowers/plans/2026-09-19-plugin-hardening.md` (1,759 lines of implementation plan — a process artifact rather than repo documentation) and the older `docs/specs/2026-09-18-…` spec this supersedes.

You offered to take this in a smaller step and I'd like to: **land the trust store and the hash-over-the-whole-plugin-directory first, with execution still disabled**, then the provider on top. Happy to review that as soon as it's up.

vinzdg:
@8nevil8 — no rush, but a heads-up: `main` moved today and this now conflicts in `Sources/Settings/SettingsView.swift` and `Tests/PreferencesTests.swift` (#256's menu-bar limits landed in both). Worth a rebase whenever you next pick it up.

The two blockers from my review still stand — the approval store being writable by the same process that drops the plugin, and the hash covering the interpreter rather than the script. My offer holds: land the trust store and the whole-directory hash first with execution still disabled, and I'll review that quickly.

8nevil8:
@vinzdg — rebased on today's `main` (custom endpoints and the plugin glyph path merge without either losing anything), and the branch now addresses everything in both comments. The PR body is rewritten to match; the short version of the two blockers:

**1. Approval store.** Moved to the keychain as you asked, then tested against exactly your threat model — and the login keychain does not hold. Any same-user process can delete another app's item without a prompt and plant its own under the same name with `security add-generic-password -A`, which the app then read silently; a provenance check over the item's partition ACL was built next and turned out forgeable too, since every ACL entry is authored by the client that creates the item (securityd stores a forged `teamid:` as given). So approvals now live in the **data-protection keychain**, whose access group comes from the caller's entitlements and which no process outside your team can reach. The one thing I need from you: that requires `keychain-access-groups` (`$(AppIdentifierPrefix)com.vinz.codenotch`) on the app and, for Developer ID, a provisioning profile that grants it. I have not touched `project.yml`, because without the profile the app would not launch and I have no identity to test with. Until it is wired, a build keeps no approvals between launches and re-asks — fail closed, with a log line saying why. Happy to add the entitlement to the project if you tell me the profile is in place.

**2. The hash and what runs.** The hash covers the whole plugin directory by content. `exec.path` and `signIn.run[0]` must be inside it or one of `/bin/sh`, `/bin/bash`, `/usr/bin/osascript`; `zsh`, `python3`, `env` and `open` are refused because each reads something user-writable to decide what to run (reproduced under the child's environment). Every argument must resolve inside the directory, symlinks may only point back inside it, command-line words are printable ASCII, and the plugin is re-inspected before every run and sign-in. A user-owned binary outside the tree is refused, which is why the CodeMie installer now writes an in-tree wrapper.

The rest of your list: *Revoke…* on the row and on directory deletion; `isPlugin` on `ProviderSnapshot` with a mark in the ring, a capsule in the tooltip and reset card, and "(plugin)" in notifications; confusable-skeleton name comparison; payload bounds; ids namespaced as `plugin-<slug>`; and the collision set is live — it refreshes when custom endpoints change and the registry rescans. The plan and the superseded spec are gone, and so is the remaining design doc; the guarantees and their limits are stated in `docs/design/plugin-protocol.md`.

On the smaller first step: I can split this into "trust store + whole-directory hash, execution disabled" if you would still rather review it that way. Say which and I will open it.

PR #276, opened 2026-09-19
Slide the notch along its edge with Alt+drag, as the Mac's ⌥-drag does

### LINKED ISSUE #251: fix: a plain drag on the Windows notch moves it, so a slipped ring-click relocates it
### Bug Description

On Windows a plain left-press anywhere on the pill, plus 4 px of movement, starts a free drag of the whole notch window (`windows/codenotch/ui/notch.html`, the `mousedown`/`mousemove` pair around line 933). On release the monitor under the window's centre takes it, the nearest of that monitor's four edges is chosen, and the position along it is written to the config (`main.rs`, the drop branch of the drag thread around line 438).

That gesture landed in #234, when snapping to an edge was the only way to change edges. Two things have happened since:

1. **#244 made a click on a ring refresh that provider.** A press that slips past 4 px is a drag, so trying to refresh a ring can relocate the notch instead. The two gestures now share the same button on the same surface with no modifier between them.
2. **#246 added the move handle and the edge drop zones**, whose whole job is choosing an edge. The free drag does that as well, from anywhere on the pill.

### Steps to Reproduce

1. Run the Windows build with the notch on any edge.
2. Press a ring to refresh it, and let the pointer slip more than 4 px before releasing — a normal mouse click on a 56 px target does this easily.
3. The notch is dragged and snaps to whichever edge it ended up nearest. The refresh never happens.

### Expected Behavior

A click on a ring refreshes that ring, whether or not the pointer moved a few pixels while the button was down. Moving the notch is a deliberate gesture of its own.

For reference, the Mac keeps the two apart in three ways:

- The drag needs ⌥ held (`Sources/Notch/NotchPanel.swift`, the `modifierFlags.contains(.option)` guard), so a plain drag on the pill does nothing and a click stays a click.
- It moves along one axis: `model.alongOffset += model.edge.isVertical ? dy : dx` (`NotchWindowController.dragged`). It never changes the edge — the handle is the only thing that does.
- The offset is saved per edge (`preferences.offset(for: edge)`) and applied as a nudge either side of centre, clamped to the screen.

### Affected Provider (if applicable)

UI / Notch Presentation

### Environment & macOS Version

Windows 11 (build 26200) — the Windows port under `windows/codenotch`

### Codenotch Version

1.13.1 (`bcd063c`)

### Additional Context

Three ways to fix it, in descending order of how much they change:

1. **Full parity with the Mac:** Alt+drag, along the current edge only, a per-edge offset from centre. A plain drag stops moving the notch, which makes the handle the one way to change edges.
2. **Keep the plain drag but constrain it to the current edge,** so it can no longer re-pick one. The slipped click still nudges the notch a little, but it cannot fly to another edge.
3. **Narrowest:** only stop a press that began on a ring from becoming a drag, and leave the gesture otherwise alone.

Happy to open the PR for whichever you prefer. Option 1 is the one that makes the handle worth having, but it takes away something that works today, so it seemed worth asking before writing it.

RawJat:
This takes the first option in #251: full parity with the Mac's ⌥-drag.

**What changes.** The drag now needs Alt held; without it every press on the pill is a click, so a click on a ring that slips refreshes it instead of carrying the notch off. With Alt, the notch follows the pointer along the edge it is on and nowhere else, inside the work area, and never changes edge — the move handle from #246 and the Edge picker in Settings do that. It is `NotchWindowController.dragged` on the Mac, which only ever moves `alongOffset`.

**Each edge remembers its own place**, as the Mac keeps `notchOffset.<edge>`. A move to another edge lands where the notch was last left there, centred if it has never been slid along it, and Recentre centres only the edge it is on. A carry used to land centred on purpose, because every edge shared one position and a carry dragged the old edge's fraction along; with a place per edge that reason is gone.

The place is stored as a fraction of the edge rather than the Mac's points (`notch_along`, 0.5 = centred). That lets an existing `notch_y` become the place for the edge it was saved on without a screen to convert it, and keeps the notch in the same part of the edge across a resolution change. `notch_y` is read once and never written back.

**One thing is lost, and it is worth saying plainly:** dragging no longer takes the notch to another monitor. The Screen picker in Settings does, as on the Mac, and its caption no longer claims the drag can.

**Settings text.** The drag hint is the Mac's own sentence — "Hold ⌥ and drag the notch to slide it along its edge. Each edge remembers where you left it." — with its ru, zh-Hans, zh-Hant, ja and uk translations taken from `Localizable.xcstrings`, ⌥ read as Alt. The Mac has no Korean, so that sentence is new and worth a native eye. The Ukrainian hint had been keyed to an older English sentence and never showed; it does now.

**Testing:** `cargo test` 103 pass, 0 fail, 2 ignored — four new: each edge keeps its own place, the old position migrates to the edge it was on, `notch_y` is never written back, and a slid notch lands exactly where it was let go (`along_at` inverts `edge_origin`). Clippy unchanged. All three pages parse.

Checked by hand on Windows 11: a plain press or a slipped click on a ring never moves the notch and still refreshes it; Alt+drag slides it along its current edge only and stops at the ends of the work area; a notch slid along one edge and carried to another lands centred there, and returns to where it was left when carried back; Recentre centres only the edge it is on; and the drag hint and Screen caption read the new wording.

**Overlaps.** #269 moves this code to `desktop/` and rewrites `drag_begin` around `platform::drag_slides_along_edge()`, so one of the two will need rebasing onto the other; its Linux drag already slides along the edge, so the intent is the same. #274 adds a Russian translation for the Screen caption this shortens, so whichever lands second needs to re-key that one string. #239, #257 and #210 also touch `main.rs`, in other functions.

Closes #251


vinzdg:
Merged as `6c2755e` — this was the cleanest of the four notch PRs. The `notch_y` read-once migration, `#[serde(default)]` on `notch_along`, and the `edge_origin`/`along_at` inverse test all held up, and the drag loop staying on its own thread is the right call. Thanks @RawJat.

One thing for a later tidy, not a problem: `drag_enabled` in `config.rs` is now referenced nowhere outside that file, and `notch_y` gaining `skip_serializing` means a downgrade to 1.15 or earlier loses the saved position and re-centres. I'll note the latter in the release notes.

PR #279, opened 2026-09-19
windows: read every Claude account, not just ~/.claude

lucaedulima:
## Summary

On Windows the usage adapter reads its credential from `~/.claude` and nowhere else, so a machine that keeps a second account in `~/.claude-<slug>` only ever sees the default profile. When that profile is the stale one, the card shows `needsAuth` while the account actually being spent stays invisible. The Mac app already reads several accounts; this brings the port in line, using the layout `CLAUDE_CONFIG_DIR` already selects rather than a new setting.

It also fixes a renewal bug that this depends on. `run_renewal` inherited the environment and cleared only `CLAUDECODE` and `CLAUDE_CODE_*`; `CLAUDE_CONFIG_DIR` is neither, so a Codenotch launched from a shell pointed at another account renewed *that* account's token while the account on screen stayed expired. On my machine that was exactly why a token expired for two days never came back: with the directory passed explicitly it renewed on the first tick.

## Changes

- `usage.rs`: `Profile` / `profiles()` discover `~/.claude` first, then every `~/.claude-<slug>` that holds a credential. A secondary directory without one is ignored, so a half-made profile never shows up as an account waiting to be signed in.
- `usage.rs`: per-account poll, renewal and back-off (`Account`), aggregated into the single `UsageSnapshot` the UI already consumes. The status is the best news any account has, so a signed-out second account cannot dim a first one that just answered; the note names the account it is about; the soonest back-off deadline is the one waited out.
- `usage.rs`: a secondary account's window ids take an `@slug` suffix, which keeps `by_id("session")` meaning the default account's session, so the notch reads exactly as before.
- `usage.rs`: `read_credentials(dir)`, and the credential now also carries `subscriptionType`, used for the account heading (`default · max`, `work · pro`) because the plan is what tells two accounts apart at a glance.
- `usage.rs`: `run_renewal(cli, dir)` sets `CLAUDE_CONFIG_DIR` explicitly instead of inheriting whatever launched the app.
- `usage.rs`: `probe_credentials()` reports every account, for `doctor`.
- `watcher.rs`: watch roots follow the same list, so a session run under a secondary profile is no longer invisible, which also held the usage poll at its idle interval while that session was live.

Each account's windows carry its name in `group`, so the card stacks them with the grouping it already draws for Antigravity's model families and no UI file changes. With a single account the reading is unchanged (ids untouched, `group` left `None`), which `one_account_reads_exactly_as_before` pins.

## Test Plan

- [x] `cargo build --locked`, `cargo test --locked` and `cargo clippy --all-targets --locked` green (the `Windows` workflow)
- [x] `Windows Package`: NSIS installer built, installed, `doctor` run, uninstalled
- [x] Tested on Windows 11 (26200) with two real accounts, one `max` in `~/.claude` and one `pro` in `~/.claude-<slug>`: both appear on the card under their own heading, the notch keeps reading the default account, and the default account's expired token renewed itself on the first tick
- [x] 7 unit tests added: account heading, id suffixing, persisted windows handed back to the account that produced them, aggregated status, soonest back-off, discovery order
- [ ] Tested on macOS: not applicable, this touches `windows/` only

Not a visual change: the card reuses the existing `group` box that Antigravity's lanes render. The resulting reading, with my account name and numbers redacted:

```json
{"status":"ok","windows":[
  {"id":"session",         "group":"default · max", "label":"Current session"},
  {"id":"weekly_all",      "group":"default · max", "label":"Weekly (all models)"},
  {"id":"session@work",    "group":"work · pro",    "label":"Current session"},
  {"id":"weekly_all@work", "group":"work · pro",    "label":"Weekly (all models)"}]}
```

---

## Second commit: one notch cell per Claude account

The first commit put the accounts on the card, but the pill kept reading the default account alone, so a machine whose default account is the idle one shows a ring at 0% while the account actually being spent is only visible on hover.

The reading already carries everything needed, so this is a front-end change in `ui/notch.html`. `claudeCells()` turns the single Claude entry into one cell per account, each holding that account's windows under the ids they would have had on their own, so `headlineOf` and `weeklyOf` need no rule about accounts at all. The cell's `id` (`claude@slug`) is what keeps the DOM unique; a new `base` field is the provider it belongs to, which is what the glyph, the activity arc, the sign-in copy and the slot choice follow. With one account, or with a reading whose windows carry no heading, `claudeCells()` returns exactly the one cell it always returned.

Verified on a two-account machine: both cells render with their own ring and percentage, the default account stays leftmost, and with `weekly_ring` on, an account idle this session but at 70% of its week reads as an amber outer ring rather than a bare 0%.


vinzdg:
Thanks @lucaedulima — the shape of this is good. The per-account `Renewer`, pinning `CLAUDE_CONFIG_DIR` per account, and bounding the home scan to `.claude-*` directories that already hold a credential (with `flatten()`/`if let Ok` throughout, so an unreadable home can't panic) are all right. I checked specifically that no token *value* is logged — `usage.rs:230` logs `token.len()` only, which is the correct instinct.

Two things to fix before it goes in:

1. **A second account's ring can never be refreshed.** `notch.html:1034` sends `p.id` (e.g. `claude@work`) to `refresh_ring`, but `main.rs:565-580` has no arm for that and returns false. Pass `p.base`.
2. **`p.name` is rendered unescaped.** `notch.html:838` does `ui().title(p.name)`, and `p.name` is now built from the home-directory slug plus `subscriptionType` read out of `.credentials.json` (`usage.rs:207`, `:536`) — both user-influenced. Escape it.

Two smaller ones:

- `refresh_provider` now gates on the *soonest* backoff across all accounts, so one rate-limited account blocks manual refresh of every other.
- A secondary account with zero windows collapses back to a single cell showing both accounts' windows grouped together.

Also a heads-up on conflicts: #257 landed today and inserts `claudeAuthHtml` immediately above the `if(p.id==='claude')` line in `renderCard()` that you change to `p.base`, and #210/#274/#276/#287/#288 have all moved `notch.html` and `main.rs` too. Worth rebasing before the next round — the `providers()` region in particular.

vinzdg:
Merged as `779b43c`, with my fixes in `d1d9a2d` and `1266ded`. Thanks @lucaedulima — the per-account `Renewer`, pinning `CLAUDE_CONFIG_DIR` per account, bounding the home scan to `.claude-*` directories that already hold a credential, and logging `token.len()` rather than the token are all right, and the `group` decoration stacking like Antigravity's model families is the correct shape.

**The merge was not mechanical, so please check I read your intent correctly.** #257 landed a few hours before this and rewrote the same polling loop for the sign-in flow. I took your per-account structure as the base and re-applied #257's four hooks onto it:

- `maybe_renew` takes the sign-in gate (`try_acquire`), so a manual sign-in and a background renewal can't rotate one credential at once.
- The tick loop skips while a sign-in is open — polling through it reads a file being rewritten and reports a signed-out account mid-login.
- A good reading clears the sign-in card's message.
- **Kept 401 and 403 separate.** Your version folded 403 into `NeedsAuth`; #257 deliberately split them, because a 403 is access denied and saying "sign in again" for a still-valid login sends people round a pointless loop. 401 still re-reads and retries once.

I also kept the empty-token guard from the old `credential_from_json` (an empty `accessToken` is signed out, not expired — it now falls through to the next candidate file), and keyed the sign-in row on `p.base` so a second account gets it too.

Two fixes to your own changes:

- **`refresh_ring` was sent the cell id.** A second account's cell is `claude@work`, which no Rust command answers to, so its ring could never be hand-refreshed — the press animated and timed out. It now resolves the base before invoking.
- **The card title was unescaped.** `p.name` is no longer a constant: it's built from the home directory's slug and `subscriptionType` out of `.credentials.json`.

One more I caught only in CI: your file restored `find_cli` to private, which #257's `claude_auth` calls (`1266ded`).

Two things I left alone, worth a follow-up if you're willing:

- `refresh_provider` gates on the *soonest* backoff across accounts, so one rate-limited account blocks manual refresh of every other.
- A secondary account with zero windows collapses back to one cell showing both accounts' windows grouped together.

CI green on all four workflows, and #257's own UI test (escaping, busy guard, refresh/backoff) passes against the merged result.

PR #282, opened 2026-09-19
Windows: add Z.AI Coding Plan usage

stanthewizzard:
## Summary
- add Windows support for Z.AI Coding Plan usage through existing ZCode, OpenCode, or manual `glm.json` credentials
- show Z.AI in the notch, tray, settings, diagnostics, and provider refresh flow
- migrate an existing customized notch selection once so the new provider can appear
- widen the horizontal Windows notch to fit six provider rings

## Scope
Windows implementation only (`windows/codenotch`). macOS is unchanged.

## Security
Credentials remain local and are not persisted by Codenotch. Encrypted ZCode tokens are skipped, and requests are limited to trusted Z.AI/BigModel hosts.

## Validation
`cargo test --release --locked -p codenotch`: 99 passed, 0 failed, 2 ignored.


vinzdg:
Thanks @stanthewizzard — the endpoint, the unit/number window identity, the ms `nextResetTime` handling, the console selection and the `enc:v1:` skip all match the Mac provider correctly. Two things block it, both where it diverges from the Mac:

**1. The `Bearer` prefix will 401 for everyone.** `glm.rs:233` sends `Authorization: Bearer <key>`. The Mac deliberately does not — from `GLMProvider.swift`: *"The monitor takes the key raw — no 'Bearer' scheme. Prefixing it is exactly what an auth failure looks like from here."* Send the key raw.

**2. Upstream prose reaches the UI and the log.** `glm.rs:354-355` puts Z.ai's `msg` into the note and into the app log. We removed exactly this on the Mac a couple of releases ago — a provider's own error text can carry account details, and it ends up in log files people paste into issues. Show the `code` only, and a fixed string for anything else.

Two more worth fixing while you're in there:

- No `glm` branch in `ring_window` (`main.rs:639-648`) or in `headlineOf`/`weeklyOf` in `notch.html`, so `glm` falls through to the Antigravity lane picker and the ring shows the *tightest* window instead of the session one. The Mac sets `headlineID: "session"`, `weeklyID: "weekly"`.
- It drops the Mac's first credential source — Claude Code's `~/.claude/settings.json` (`ANTHROPIC_AUTH_TOKEN` plus a Z.ai `ANTHROPIC_BASE_URL`) — and invents a `glm.json` the Mac has no notion of. Worth matching the Mac's source order so the same machine works on both.

Nit: status `"none"` at `glm.rs:361` isn't a status the UI knows.

Note that #276, #287, #288, #274, #210 and #257 have all landed since you opened this, so `config.rs`, `main.rs` and both HTML pages have moved — you'll want a rebase. Happy to look again after.

vinzdg:
Merged as `174a5c7`, with my fixes in `cd3f2a1`. Thanks @stanthewizzard — the endpoint, the unit/number window identity, the ms `nextResetTime`, the console selection, the `enc:v1:` skip and the one-time GLM migration were all correct, and I took them as written.

What I changed rather than sending back:

- **The `Bearer` prefix.** `GLMProvider.swift:110` is explicit that the monitor takes the key raw and that prefixing it is exactly what an auth failure looks like from here — every read would have answered 401. Fixed the module doc too, which described it as Bearer-authenticated.
- **The upstream prose.** Z.ai's `msg` went into both the card and the log; now it's the code only.
- **Added Claude Code as a source.** `~/.claude/settings.json` `env.ANTHROPIC_AUTH_TOKEN` with a Z.ai `ANTHROPIC_BASE_URL`, claimed only when the base URL names a Z.ai console. It's the Mac's first source and the documented way to point Claude Code at the plan — without it, a machine set up the normal way showed nothing on Windows while working on the Mac. Your `glm.json` stays as the first source.
- **Ring routing.** `glm` had no arm in `ring_window`, `headlineOf` or `weeklyOf`, so it fell through to Antigravity's lane picker and showed the tightest window rather than the session. Now `session`/`weekly`, matching the Mac's `headlineID`/`weeklyID`.
- Dropped a `"none"` status the UI has no case for.

Merge note: today's six Windows merges moved the notch to 44 px rings, so your six-ring width calculation resolved to 6 × 44 rather than 6 × 56. The flat notch needs 504 px and has 650.

CI green on all four workflows.

PR #283, opened 2026-09-19
Windows: add edge-notch auto-hide

stanthewizzard:
## Summary
- add an auto-hide option for the Windows edge notch
- retract the notch to an 8 px tab after the pointer leaves and reveal it on hover
- support left, right, top, and bottom edges
- hide the move/settings handles and decorative fillets while retracted, preventing the visual artifacts outside the tab
- expose Auto-hide / Always visible in Windows settings

## Scope
Windows implementation only (`windows/codenotch`). macOS is unchanged.

## Validation
`cargo test --release --locked -p codenotch`: 100 passed, 0 failed, 2 ignored.


vinzdg:
Closing — #288 has landed and covers edge auto-hide.

Two things here would have needed fixing first, recording them in case they're useful elsewhere:

- `ui/notch.html:27,917` — retracting animates `#pill` with a 160 ms `transform` transition, then `setRetracted` measures via `requestAnimationFrame(reportHot)` one frame in. `getBoundingClientRect` reflects the in-progress transform, so the hot rect reported for a retracted notch is close to the full pill: click-through stays off over a large invisible region and the 8 px tab is never the real reveal target. #288 sidesteps this by measuring the static `#rest` element.
- `src/config.rs:133` — `default_auto_hide() = true` means every existing config starts auto-hiding on update, which the test asserts as intended. #288 reads the raw JSON to tell a fresh install from an upgrade so existing notches stay open.

Also nothing suppressed retraction during `begin_move`, `drag_begin` or while the context menu was up, so the pill vanished mid-carry, and the new strings went in via bare `.textContent =` rather than `ui(...)`, so they'd have been English in every locale.

Thanks @stanthewizzard — and #282 (Z.AI) is still open and worth finishing; I've left review notes on it.

PR #287, opened 2026-09-19
Carry the Windows notch across monitors by its handle

RawJat:
The move handle only ever offered the edges of one monitor. `begin_move` took the notch's screen once, read every target against it, and on release wrote the edge alone. A pointer carried onto a second monitor still picked an edge of the first, and the overlay stayed behind — drawn there at the wrong size, since its figures belonged to the other screen.

**The zones now follow the pointer.** Whichever screen it is over, the overlay moves there, the edge is read against that screen's work area, and letting go writes the screen as well as the edge. The right edge of another screen counts as a move too, though the edge has the same name. Over the gap a shorter screen leaves beside a taller one, the carry keeps the screen it was last over rather than flicking between them.

**Why now:** since the Alt+drag change (#276), the drag no longer leaves its own edge, so this becomes the one gesture that takes the notch to another monitor; the Screen picker in Settings is the other way. The Mac's handle is single-screen — `beginMove` takes `currentScreen()` once — so this goes past it rather than porting it.

**Scale.** The overlay is pinned in physical pixels, and pinned again if Windows resizes it for a new DPI after the move, the way `place_notch` pins the notch. The silhouette is in logical pixels, so it keeps its size on a screen at another scale — the size the notch will land at. Screens are told apart by where they are rather than by name, which the platform is not obliged to report.

**Moving between scales without it showing.** Crossing onto a monitor at another scale, Windows first resizes a window by the ratio of the two scales, and the fix-up above then puts it right — which is correct, but played out on screen: the zones shrank and grew again on every crossing, and a notch landing on the other monitor visibly jumped sizes. So the overlay is hidden for the crossing and shown again once the new zones are drawn. The overlay is also made unfocusable, like the notch, since it is shown again on every crossing.

The notch needs more care, because hiding its window does not work: a hidden WebView2 stops painting and throttles its timers, so the page re-zoomed itself for the new pixel ratio only once shown again, in plain view. Instead the window stays up and the page empties itself — the window is transparent, so that is an invisible notch — while the WebView keeps doing its layout. It is revealed on the first `report_dpr` from the end of a burst of resizes that needs no further correction, i.e. once the layout has actually stopped changing, with a 700 ms fallback so it can never stay hidden. A landing at the same scale is not hidden at all, and neither is any other placement.

**Testing:** `cargo test` 104 pass, 0 fail, 2 ignored, including a new test for which screen a carry is over, the shared border between two, and the gap beside a shorter one. Clippy unchanged.

**Depends on #276** (Alt+drag): both touch the landing in `begin_move`, so this is built on top of it and should go in after. #269 moves `dropzones.rs` and `main.rs` to `desktop/`; #239, #257 and #210 touch `main.rs` in other functions.


vinzdg:
Merged as `107452b`. The DPI handling is sound and emptying the page rather than hiding the window is the right trick for WebView2 — thanks @RawJat.

I pushed one follow-up on top, `7121fb8`: `c.notch_monitor = mon.name.clone()` wrote `None` when Windows wouldn't name the destination display, and `None` doesn't mean "this screen" — `target_screen` reads it as "the primary". So carrying onto an unnamed monitor was accepted and saved, then the notch left it at the next placement. Since the Settings picker already lists unnameable screens disabled for exactly that reason, the carry now keeps the edge you chose and leaves the notch where it was rather than record a move that won't survive.

Two smaller things I left alone: the 40 ms `sleep` inside the 16 ms carry poll in `dropzones.rs:101` stalls pointer tracking on each screen crossing, and `same_screen` compares geometry so mirrored displays are indistinguishable. Neither is worth a change on its own.

PR #288, opened 2026-09-19
Fold the Windows notch to a pill when not in use, as the Mac's Show on hover does

RawJat:
The Mac's Show setting has three modes and Windows had two. The one missing was the Mac's default, **Show on hover**: the notch rests as a small pill at the edge and opens the moment the pointer reaches it.

**The fold.** Folded, the notch is the Mac's resting pill, `NotchLayout.pillWidth` × `pillHeight` — 10 × 79 here, in the same unscaled measures the orb uses. The open pill keeps its whole layout and a `clip-path` draws it down to that, so everything measured off it (the card's anchor, the handles, which ring is under the pointer) stays true while folded. A thin outline arrives with it, since a black pill on a black wallpaper is otherwise nothing. The rings leave by fading and sliding toward the edge one after another, the Mac's `NotchMotion.stagger`, and the handles fade with them.

**When.** It opens on contact and folds 0.45 s after the pointer has gone (`foldGrace`). What wakes it is the pill's length along the edge and its depth plus the Mac's `pillHotZone`, 34 px, into the screen. The pointer watchdog reports the crossing from the transition that already switches click-through, because once click-through is back on the page is sent nothing. It never folds mid-carry, mid-drag or under the right-click menu, which all take the pointer off the notch on purpose.

**Keep open** in the right-click menu is the Mac's own shortcut (`onToggleKeepOpen`): it switches between Always show and Show on hover, the same setting from another place, and an open Settings window follows it.

**Defaults.** A fresh install starts on Show on hover, as on the Mac. A config saved before the setting existed keeps the always-open notch it had, so nobody's notch starts folding on an update. Settings gains the third button with the Mac's wording and its ru, zh-Hans, zh-Hant, ja and uk translations; the Mac has no Korean, so those strings are new and worth a native eye.

**The rings take the Mac's share of the body, 44 in 70.** With the weekly ring set to Outside, the outer ring came within 2.8 px of the screen edge. The Mac keeps it "far enough in that the bezel is never touched", and has room to because its ring is 44 pt in a 70 pt body; ours was 56 in 70. The arcs scale with the ring from their viewBox, the glyphs take the same factor, and the outside ring now sits about 10 px clear, as on the Mac. The percentage labels keep their size.

**Testing:** `cargo test` 105 pass, 0 fail, 2 ignored, including a new test that only a fresh install starts on hover. Clippy unchanged. All three pages parse.

**Alongside #283 and #239.** #283, opened yesterday, reaches for the same thing, and #239's Auto-hide is nearby. The differences are in how closely each follows the Mac: this one folds to the Mac's own resting pill rather than sliding the whole pill off the edge to a strip; it adds the Mac's third mode to the existing Show row rather than a second row beside it; it starts on hover only for a fresh install, so an update does not change a notch someone is already using; it keeps the Mac's guards against folding mid-carry, mid-drag or under the right-click menu, and its Keep open; and it carries the Mac's translations. They will conflict in the same places, so it is one or the other — happy to bring anything across from either, or to fold this into theirs, whichever you would rather keep.

**Depends on #276 and #287**, and is built on top of both; it should go in after them. #269 moves these files to `desktop/`, and #274 adds Russian to `settings.html`.

## Preview

https://github.com/user-attachments/assets/ed80a1b4-96c4-45da-b5ea-16e81843da47


vinzdg:
Merged as `faf85b2`. This is the best parity of the three auto-hide attempts: a third mode on the existing **Show** row rather than a second control, the Mac's `foldGrace`/`pillHotZone`/stagger, `keep_open` in the notch menu, and guards against folding mid-carry, mid-drag and under the menu. `keep_open_on_upgrade` reading the raw JSON to tell a fresh install from an upgrade is a nice touch — existing notches stay open, which is what I'd want. Thanks @RawJat.

I'd flagged `f9d5b27` (rings 56→44 in 70) as unrelated scope and was going to strip it. I checked it before doing so and kept it: `Sources/Notch/NotchLayout.swift:52` has `ringDiameter = Design.px(117) // 44pt, the design spec's anchor`, with `sideRingMargin` derived from it, so your ratio is the Mac's and the 2.8 px bezel clearance on the outside weekly ring was a real defect. Your reasoning was right and mine was wrong.

Left as-is: the `mousemove` handler in `notch.html` sets `pointerIn` from a DOM event while Rust's `notch_pointer` is the authority. Harmless today because Rust always emits the `false` transition, but two owners of one flag is worth collapsing next time you're in there.

PR #289, opened 2026-09-20
feat: add custom endpoints support with budget and icon customization

mulhamna:
### Summary
* Provide the ability to add custom OpenAI/Anthropic/generic API endpoints alongside existing built-in providers.
* Allow tracking estimated usage or spending against configurable monthly budgets with real-time status rings.
* Support custom endpoint icons via built-in presets or user-uploaded images.

### Changes
* **Model & Storage**:
  * Added `CustomEndpoint` and `PricingRule` models to store custom endpoint configuration, auth headers, token pricing, and budgets in `UserDefaults`.
  * Added `CustomIconStore` to store and manage uploaded custom endpoint icons securely in Application Support.
* **Provider & Polling**:
  * Implemented `CustomEndpointProvider` and probe network logic to check models/usage or track estimated budget spend.
  * Dynamically synchronized `customIconFilename` across `UsageProvider`, `ProviderSnapshot`, and `UsageStore`.
* **Notch & UI**:
  * Updated `ProviderRing` and `ProviderGlyphView` to render custom endpoint icons smoothly in notch rings and tooltips.
  * Clamped tooltip tail offsets within `TooltipSilhouette`, `TooltipShell`, and `UsageResetCard` to prevent seam distortion near card corners.
* **Settings**:
  * Added dedicated Custom Endpoints management tab in Settings with full CRUD, test connection probing, preset picker, pricing calculator, and icon upload.
* **Tests**:
  * Added unit tests for custom endpoints, pricing calculation, and status snapshot generation.

### Test plan
- [x] Configure a custom endpoint with token pricing and monthly budget.
- [x]  Configure a custom endpoint with https://vikey.ai
- [x] Verify custom icon upload and preset selection display properly in settings.
- [x] Verify notch ring renders custom icon and budget percentage accurately.
- [x] Verify hovering notch card shows clean tooltip silhouette without tail distortion or bottom gap.
- [x] Run automated test suite (`xcodebuild test`) — all tests passing.

## Evidance
Screenshot 2026-09-20 at 07 34 32
Screenshot 2026-09-20 at 07 51 03


vinzdg:
Thanks @mulhamna — this is a lot of careful work, and the Settings pane, the icon plumbing and the tooltip tail-clamp are all reasonable. But I can't take the provider core as it stands. Blockers:

**1. API keys go into UserDefaults in plaintext.** `CustomEndpoint.swift:34` puts `apiKey` on a `Codable` struct and `Preferences.swift:59-66` JSON-encodes the lot into defaults — a plist readable by any process running as the user, and swept up by backups and sysdiagnose. Every other credential in this app is read from the Keychain or from another tool's store, and `Sources/Providers/KeychainItem.swift` already exists. Keys need to go there with only a reference in defaults.

**2. `baseURL` is never validated to be http/https.** `URL(string:)` accepts any scheme, and `account().manageURL` is handed straight to `NSWorkspace.shared.open(url)` (`SettingsView.swift:2031-2043`). A user- or preset-supplied scheme gets launched. Validate the scheme at save time and drop `manageURL` for anything non-https.

**3. The pricing calculator is dead code.** `promptTokensUsed` and `completionTokensUsed` are never incremented anywhere in the PR — only zeroed at `CustomEndpointsSettingsView.swift:561` — and `lastCheckedAt` is never written. The ring only moves if someone hand-types "Direct Spend".

**4. `fetchSnapshot` does no network at all** and `CustomEndpointProvider.swift:229` hardcodes `status: .ok` regardless of reachability. CONTRIBUTING's "never invent a number" covers this.

**5. Data race.** The `@Sendable` closure reads `@MainActor` `Preferences.customEndpoints` from the provider actor and from `nonisolated` `glyph`/`displayName`/`account()` (`CustomEndpointProvider.swift:104-124`). It only compiles because strict concurrency is `minimal`. The pattern here is `nonisolated static func storedX(defaults:)` — see `Preferences.swift:466-512`.

Worth fixing too: `AppDelegate.swift:175-183` re-registers every custom provider on each `` mutation (so every Test click causes a refresh storm and ring flicker), `isEnabled` is never consulted so disabled endpoints still poll, "Monthly Budget" never resets (`resetsAt: nil`, so spend accumulates forever) and `Int(remaining)`/`Int(spend)` truncate dollars, and roughly 45 new `L10n.t` keys never reach `Localizable.xcstrings` — plus some hardcoded English bypassing `L10n.t` entirely at `CustomEndpoint.swift:11-18` and `CustomEndpointProvider.swift:154/162/185/194`.

One concrete suggestion: **split the tooltip tail-clamp changes out now.** The `TooltipCard.swift` / `UsageResetCard.swift` work looks correct and independent of all of the above, and I'd merge that on its own this week while the endpoint core gets another pass.

mulhamna:
Thanks a lot for the detailed and sharp review @vinzdg, really appreciate you taking the time to go through this!

I have addressed all the blockers and feedback, and restructured the branch into 2 clean commits so the tooltip tail-clamp fix is isolated in the first commit (in case you want to cherry-pick or review it on its own):

1. **API Key Security (Keychain)**:
   * Removed `apiKey` from `CustomEndpoint`'s `Codable` serialization entirely — `UserDefaults` now only stores endpoint metadata and references.
   * Credentials are now stored and managed securely in the macOS Keychain via `KeychainItem` (`saveAPIKey` / `deleteAPIKey`). Any legacy keys in defaults are migrated to Keychain on load.

2. **URL Validation & Safe `manageURL`**:
   * Added `CustomEndpoint.isValidURL` validation checking for `http`/`https` and a valid host before saving.
   * `account().manageURL` is strictly limited to `https` schemes to avoid opening arbitrary or unsafe schemes.

3. **Pricing & Budget Tracking**:
   * Removed the dead-code token calculator and token price inputs since Codenotch doesn't proxy requests.
   * Refined it to a clean Monthly Budget & Spend Tracking model with automatic monthly reset dates (`resetsAt` pointing to the 1st of next month) and exact dollar/cents breakdown without integer truncation.

4. **Live Network Probing & Provider Status**:
   * `fetchSnapshot()` now performs a real network probe against the endpoint (`/models` or base URL) to verify reachability and authentication (throwing `needsAuth` on 401/403 and `badResponse` on reachability errors). No hardcoded status.

5. **Concurrency & Thread Safety**:
   * Replaced the `@Sendable` closure reading `@MainActor Preferences` with `Preferences.storedCustomEndpoints(defaults:)` following the `storedX` pattern.
   * In `AppDelegate`, provider registration is now filtered by `isEnabled` and deduplicated by enabled IDs (`.removeDuplicates()`), eliminating refresh storms and ring flicker when testing or updating fields.

6. **Localization**:
   * Added all missing keys (49 strings) to `Localizable.xcstrings` and wrapped remaining hardcoded texts with `L10n.t`.

All 1,646 tests are passing. Let me know what you think when you have a moment!

mulhamna:
Added an option to track usage by **Token Count (in Millions)** in addition to USD ($), particularly useful for custom endpoints, local inference (vLLM, Ollama, LM Studio), or proxies where usage is measured in tokens rather than direct currency spend.

### Updates:

- **Tracking Unit Selector**: Added a segmented control in Settings (`USD ($)` vs `Tokens (Millions)`).
- **Dynamic Form & Formatting**:
    - Selecting tokens adjusts inputs to *Monthly Budget (M tokens)* and *Tokens Used (M tokens)*.
    - Headline and notch rings support compact token formatting (e.g. `250k`, `2.5M`, `10M`, `1.2B`) or percentage (`%`) based on the toggle.
    - Tooltip card shows remaining/used token progress and counts cleanly (`X.XM / Y.YM tokens remaining`).
- **Full Backward Compatibility**: Defaults to USD currency tracking for all existing endpoints without breaking archives or saved settings.
- All unit tests passing.

Here is a quick preview of the settings UI:
Screenshot 2026-09-20 at 23 29 14

vinzdg:
Merged as `eb8cfbc`. Thanks @mulhamna — this is a big turnaround and I want to be specific about it, because you fixed all five blockers properly rather than papering over them:

1. **Keys** now live in the keychain behind a computed `apiKey`, with `legacyApiKey` decode-only and never encoded. Service and account naming is sane.
2. **`baseURL`** is validated to http/https with a non-empty host at save time and before every probe, and `manageURL` is https-only.
3. **The pricing calculator** — you deleted it rather than making the dead token counters work, and replaced it with a manual budget that has a real `nextMonthlyResetDate()`. That was the right call, and `Double` money instead of the truncating `Int()` is a bonus.
4. **`fetchSnapshot`** does a real probe now, and `status: .ok` is only reachable after one succeeds.
5. **The data race** is gone, via `nonisolated static func storedCustomEndpoints(defaults:)` — the repo's own pattern.

The 51 catalog keys are a clean append with no reordering and every format specifier matching, which is the thing most likely to go wrong in this file.

Follow-up of mine on top, `784985e`, four things:

- **The migration wasn't finished.** It ran inside `init(from:)`, which is called on *every* decode — and `storedCustomEndpoints()` decodes the whole list on every provider property access, so a `SecItemAdd` fired several times per repaint. And the plaintext key was never removed from the plist: `customEndpoints` is assigned during `Preferences.init`, where `didSet` doesn't run, so nothing wrote the list back and the key stayed readable on disk until the user happened to edit that endpoint. The decode now only *carries* the key and `Preferences.movingLegacyKeysToKeychain` moves it once at load, rewriting the plist without it.
- **Redirects.** A 302 was re-sending the `Authorization` header to wherever the endpoint pointed. The address is the user's; where it redirects to isn't.
- **Error text.** `error.localizedDescription` was reaching the card and the saved health — a `URLError`'s description carries the host and path. Now it's the kind of failure.
- **The model list** is bounded before storage, since it's JSON-encoded into the plist and decoded on every property access.

Plus the seven `L10n` keys these paths use that the catalog was missing (including `"USD ($)"` and `"Show remaining budget instead of spent"`, which had a near-miss entry).

Two things I left for you, neither blocking:

- `CustomEndpoint.apiKey` calls `KeychainItem.read` directly on every poll, bypassing `CredentialCache`. `KeychainItem.swift:104-108` explains why own-item readers go through it — an ad-hoc rebuild changes the signing identity and *will* prompt.
- `lastCheckedAt` is declared and never written.

And the tooltip tail-clamp work is still bundled in here rather than split out. It's fine — it went in with the rest — but it's genuinely independent and would have merged days ago on its own.

1,710 tests green.

PR #296, opened 2026-09-21
Fail when a remote extra ring ignores Settings

hygorfragas:
## Summary
Adds a failing XCTest that records a gap: an extra remote reading still appears on the notch when that assistant is switched off in Settings. The CI on this repo is the intended way to observe the RED; the test is not meant to merge green yet.

## Changes
- `Tests/RemoteUsageTests.swift`: `testDisabledRemoteAssistantDoesNotPublishAnExtraRing` injects a `claude:Artemis` extra cell with `claude` disconnected and asserts it stays off the notch.

## Test plan
- [ ] CI `make test-ci` on this PR (expect `RemoteUsageTests` to fail)
- [ ] Not tested on macOS locally (this machine has Command Line Tools only)
- [ ] UI / Notch interactions not applicable (unit test only)

Made with 💻 by Hygor

Made with [Cursor](https://cursor.com)


hygorfragas (commented):
ok

vinzdg:
Thanks @hygorfragas — but I'm closing this one, and I want to explain properly rather than just decline it.

The test registers two *independent* `UsageProvider`s — `claude` and `claude:Artemis` — and asserts that switching off the first suppresses the second. Nothing in the app maps a child id back to a parent, so it fails. But it fails because it is asserting a rule the product does not have, over an entity that does not exist: there is no "remote extra ring" concept in `Sources/`, and the `parent:child` id shape is invented here.

The two real mechanisms are different from each other and from this:

- **Extra cells** are local model cells, keyed `&lt;id&gt;:model:&lt;modelId&gt;` (`Sources/Model/UsageModel.swift:278`). Those already cascade correctly, because they are expanded from the *parent's* snapshot — the parent is removed first, so the children go with it. Two independent providers is not how an extra cell is produced.
- **Multiple accounts** are `~/.claude-&lt;slug&gt;` profiles (`Sources/Providers/ClaudeProfile.swift:122`), which give ids like `claude-work`. Those are deliberately independent: they register as separate providers, get their own Settings toggle, and appear separately in `Preferences.disconnectedIDs(among:)`. Switching off `claude` is *meant* not to switch off `claude-work` — someone with a personal and a work account should be able to hide one and keep the other. Merging this test would lock in the opposite.

One practical thing regardless of the above: the PR body says the test is expected to fail on merge. `project.yml:126-130` globs all of `Tests/`, so a red test on `main` turns CI red for every other open PR too. A bug report belongs in an issue; a test belongs in the commit that makes it pass.

The test itself is well built — isolated `UserDefaults` suite with a teardown block, `@MainActor`, deterministic, no sleeps. That part matches the house style. Two small notes for next time: the `// SPECSFY: US-001 …` marker appears nowhere else in `Tests/`, and the probe reimplements `account()`, `signInRoute`, `signOut` and friends that already have defaults in `Sources/Providers/ProviderAccount.swift:90-111` — compare `Tests/UsageRefreshDeadlineTests.swift:14-37`, which stubs only `fetchSnapshot`.

If you do want a parent/child cascade — say, one assistant read over two machines — open an issue describing the behaviour you want and I'll take it seriously. It needs an id convention decided first.

PR #297, opened 2026-09-21
Breathe a provider's mark in the menu bar while it is working

ozgurasan-agent:
## Summary

The menu bar item says how much of a five-hour window is gone, but not whether anything is spending it *right now* — the only way to tell was to switch to the agent's own window. Each provider shown in the bar now pulses its mark while it is doing AI work, and keeps still otherwise. The percentage, the countdown, the rule between readings and the item's width never move.

Activity is normalised **per provider** rather than worked out in the menu bar, so the bar consumes a generic state by id and knows nothing about how any provider was read:

```
provider-specific detection → ProviderActivityState → ProviderActivityBoard → status item → pulse
```

`ProviderActivityState` is `active` / `idle` / `unknown` for one provider. `ProviderActivityBoard` merges every source that can speak for one (any source seeing work wins), drops providers that are switched off in Settings, and owns the sources that exist only for this — starting them while one of their providers is connected. Each provider's state is its own; there is no global "something is working" flag.

Sources already running are read where they are (no new polling loops): the notch's `AgentActivityMonitor`s — Claude and Codex per profile, Cursor, Antigravity, Grok, Kimi, Gemini API — plus the Ollama relay and LM Studio's socket. Nothing new reaches the notch, its chimes or the usage refresh schedule.

## Changes

- **New `Sources/Sessions/ProviderActivity.swift`** — the normalised state, the `ProviderActivitySource` protocol, an adapter for a single-provider `AgentActivityMonitor`, and the board that merges and gates by connected provider.
- **New `Sources/Sessions/CopilotActivityMonitor.swift`** — GitHub Copilot's agent writes turn boundaries into `~/.copilot/session-state/<id>/events.jsonl` (`user.message` → `assistant.turn_start` → `tool.execution_*` → `assistant.turn_end`). A turn that ends asking for tools is followed at once by another, so the work is over only when one ends with an answer. The log names no process, so the CLI's own `process-<start ms>-<pid>.log` files gate it (fail-open) and silence expires a turn whose CLI was killed.
- **New `Sources/Sessions/OpenCodeActivity.swift`** — OpenCode records the billing `providerID` and a completion time on every assistant message, so an unfinished row credits the account it spends: OpenCode Go, GLM's and Zhipu's coding plans, MiniMax's Token Plan, Kimi For Coding, Copilot, DeepSeek and Ollama's cloud (ids checked against the models.dev catalogue OpenCode uses). Deliberately **not** attributed: `zai`/`minimax` pay-as-you-go, which do not touch those plans, and `anthropic`/`openai`, which cannot be tied to one Claude or Codex profile of several. `OpenCodeGeminiActivity` is now that same reader with `google` asked for — its tests are unchanged and still pass.
- **Ollama relay** — it reported only the *thinking* phase, which a model that does not reason never enters. It now reports the whole request (start to last byte, failure or cancellation), which is what working means for the local runtime, and tells hosted models from local ones by their `cloud` tag, so `ollama-local` and `ollama` (cloud) each get their own state.
- **New `Sources/App/StatusItemPulse.swift`** — the drawing. A mask on the status button's own layer, animated by Core Animation: AppKit goes on tinting the item exactly as before (light and dark bars, a wallpaper-tinted one, the dimmer bar of an inactive display), only the mark's alpha moves, and it moves in the render server — nothing runs in the app between frames, and with everything idle there is no mask at all. Each hole is a full-height column, so it never has to know which way up the button's layer is. A mark that stops eases back to full instead of snapping. The open menu's highlight is drawn outside the button's layer and is untouched.
- **`StatusItemSummary` / `StatusItemController`** — an `isActive` flag per entry, an `activity` argument to `make`, the mark's box exposed for the pulse, and `· Working` appended to the tooltip and VoiceOver label, which is the one thing a pulse cannot tell a screen reader.
- **Reduce Motion** — the breath is replaced by a still dot in the mark's lower corner, cut clear of it and drawn into the same template image. Followed live via `NSWorkspace.accessibilityDisplayOptionsDidChangeNotification`.

### Provider coverage

| Provider | Detection | Event-driven / polling | Confidence |
| --- | --- | --- | --- |
| Claude (per profile) | existing session registry + transcript turn state | dir watcher + 2 s | authoritative |
| Codex (per profile) | existing rollout `task_started`/`task_complete` | 2 s | authoritative + recency |
| Cursor | existing `composerHeaders.unfinishedRunAt` | 2 s | authoritative |
| Antigravity | existing transcript tail + recency | 2 s | heuristic |
| Grok | existing `updates.jsonl` recency + pid | 2 s | heuristic |
| Kimi | existing `wire.jsonl` turn records (+ OpenCode plan turns) | 2 s | authoritative |
| Gemini API | existing CLI recency + OpenCode + Hermes lease | 2 s | mixed |
| LM Studio | socket instance state | 0.4 s poll | authoritative |
| Ollama (local) | **new** relay request lifecycle | event-driven (relay opt-in) | authoritative where routed |
| Ollama (cloud) | **new** relay, `cloud`-tagged models | event-driven (relay opt-in) | authoritative where routed |
| GitHub Copilot | **new** agent event log + OpenCode | change-gated 2 s | authoritative for CLI/VS Code agent sessions |
| OpenCode (Go) | **new** unfinished assistant turn | change-gated 2 s | authoritative |
| GLM | **new** OpenCode coding-plan turns | change-gated 2 s | authoritative for OpenCode; blind to ZCode and Claude Code routed at Z.ai |
| MiniMax | **new** OpenCode Token Plan turns | change-gated 2 s | same limitation |
| DeepSeek | **new** OpenCode turns | change-gated 2 s | same account in the common case |
| Devin, Kiro, Command Code, QianwenAI | **none** | — | no in-flight signal these tools write to disk could be verified; they are `unknown`, which is drawn exactly like idle |

## Test Plan

- [x] Run test suite (`make test`) — **1753 tests, 0 failures**, 43 of them new (`ProviderActivityTests`, `StatusItemActivityTests`, `CopilotActivityTests`, `OpenCodeActivityTests`, relay lifecycle tests).
- [x] Tested on macOS (version: 27.0, Xcode 27.0), both a 1× external display and the 2× built-in one.
- [x] UI / menu bar interactions verified live, by measuring the real menu bar's pixels through an opt-in check (`StatusItemActivityVisualTests`, skipped unless `CODENOTCH_ACTIVITY_VISUAL_OUTPUT` is set) plus the running app driven by a real Claude Code session.

Measured peak ink of each mark across frames of the real menu bar (min–max; a mark that pulses swings, one that does not is flat), with the digits beside them as a control:

| Phase | Claude mark | Codex mark | digits |
| --- | --- | --- | --- |
| Claude working | 0.708 – 0.936 | 0.932 flat | 0.935 flat |
| both working | 0.715 – 0.936 | 0.706 – 0.932 | 0.935 flat |
| Codex working | 0.936 flat | 0.699 – 0.932 | 0.935 flat |
| both idle | flat | flat | flat |
| Reduce Motion, Claude working | flat (dot instead) | flat | flat |
| Codex working but not chosen for the bar | flat | not shown | flat |
| limits switched off | plain 22 pt icon, nothing animates | — | — |

The same check timed the process: 0.15 % CPU with two marks pulsing against 0.3 % with everything idle — the animation costs the app nothing per frame, as the notch's spinning arc does not.

Also verified: usage refreshes and countdown ticks never change an activity state; a provider not chosen for the bar cannot appear in it; switching monitoring off in Settings takes a provider back to `unknown`; and no new usage fetch is made for activity.

Peak and trough of one breath, captured from the real menu bar — only the mark changes:

```
[✳] 80% · 4h 28m | [⊛] 100% · 1h 05m     mark at full
[✳] 80% · 4h 28m | [⊛] 100% · 1h 05m     same item, mark at 62%
```

With Reduce Motion the mark keeps still and takes a dot in its lower corner instead.


vinzdg:
Thanks @ozgurasan-agent. The pulse itself is well built and I checked it specifically because I expected the opposite: `Sources/App/StatusItemPulse.swift:97-108` is one `CABasicAnimation` on a mask layer with `repeatCount: .infinity`, running in the render server — no `NSStatusItem` image is redrawn per frame, and with nothing working there is no mask at all. **Idle wake-ups added by the animation: zero.** That is the right way to do it.

The blocker is next to it. This adds two always-on 2-second pollers — `Sources/Sessions/CopilotActivityMonitor.swift:57-63` and `Sources/Sessions/OpenCodeActivity.swift:191-197` — gated on `setConnected`, i.e. on the provider being switched on in Accounts. Not on the menu bar item existing, and not on `showsLimitsInMenuBar`. So someone running notch-only, who never opted into any of this, pays about **60 extra timer fires a minute**, and Copilot's does an unconditional `contentsOfDirectory` plus a `stat` per session every tick with no write-gate. OpenCode's narrows to two `stat`s when idle; Copilot's should do the same.

The 2-second interval is fine — it matches `AntigravityActivityMonitor.swift:27` and the others. It's the gating. These should start when something actually consumes them and stop when it doesn't.

1.13.0 was a battery release and people noticed; I'd rather not give that back for an animation that is off-screen most of the time.

Two smaller ones:

- `StatusItemSummary.swift:113-118` appends ` · Working` to `detail` on every active↔idle flip, so `next != summary` and the whole `NSImage` is rebuilt on each transition even when the artwork pixels are identical. Gate the rebuild on `badgesActivity`.
- `StatusItemController.swift:187-189` registers the occlusion observer with `object: item.button?.window`, which is `nil` at install time — so it silently becomes an all-windows observer.

Everything else is clean: timers are `[weak self]` and invalidated in `stop()`, observers torn down in `hide()`, and `ProviderActivityTests`/`CopilotActivityTests` are real tests with real fixtures.

Fix the gating and I'll take it. Note that #299 and #298 have landed, so this will want a rebase — `AppDelegate.swift` and `StatusItemController.swift` both moved.

vinzdg:
@ozgurasan-agent — keeping this open rather than closing it in favour of #313, and here's how I'd like to split them.

#313 answers the poller objection cleanly: no new timers, gated on the status item rather than on a provider being connected. I'll take that as the mechanism.

**What's valuable here and isn't in #313** is the other ~1,900 lines: `CopilotActivityMonitor`, `OpenCodeActivity`, the Ollama relay and LM Studio feeds, the `ProviderActivityBoard` abstraction, and around 880 lines of real tests. #313 only pulses providers that already have monitors on `main`. I want the new sources.

So: rebase this onto #313, drop the renderer (it's the same idea, and your `StatusItemPulse` mask is the better one — I've asked for it to move *into* #313), and move the two 2-second pollers behind `ProviderActivityBoard`'s own enable gate so they start only when something consumes them. `CopilotActivityMonitor.swift:57-63` should also narrow to a couple of `stat`s when idle the way `OpenCodeActivity` already does, rather than an unconditional `contentsOfDirectory` plus a `stat` per session every tick.

Two smaller ones from the review: `StatusItemSummary.swift:113-118` appends ` · Working` to `detail` on every active↔idle flip, so the whole `NSImage` is rebuilt on each transition even when the pixels are identical — gate that on `badgesActivity`. And `StatusItemController.swift:187-189` registers the occlusion observer with `object: item.button?.window`, which is nil at install time, so it silently becomes an all-windows observer.

For the record, I checked the animation itself and it's free — one `CABasicAnimation` in the render server, no image redrawn per frame, and no mask at all when nothing is working. The objection was only ever the two pollers beside it.

PR #298, opened 2026-09-21
Show weekly usage in the menu bar

ozgurasan-agent:
## Summary

- Add an opt-in weekly-consumption ring around each eligible provider glyph in the macOS menu bar without increasing status-item width.
- Add a default-off, immediately applied and persisted setting under Appearance → Menu Bar.
- Keep the existing five-hour percentage, reset copy, provider visibility, default-icon fallback, and status-menu quick toggle intact.
- Reuse provider snapshots and semantic `weeklyID` declarations; no extra usage polling or provider requests are introduced.

## Implementation

- Draw a monochrome clockwise progress ring inside the provider glyph box, including a faint track for a valid 0% reading.
- Resolve weekly usage through `ProviderSnapshot.weeklyLimitWindow`, preserving the notch rule that avoids duplicate headline/weekly rings.
- Omit unavailable, invalid, negative, non-finite, or expired weekly readings without hiding the provider.
- Declare Grok and QianwenAI weekly-only windows through the same provider metadata path used by Claude, Codex, and other providers.
- Include exact weekly used/remaining text in the tooltip and VoiceOver label.

## Provider Coverage

Weekly metadata is available for Claude Code, Codex, Antigravity, GLM, MiniMax, Grok, Devin, OpenCode, Command Code, Kimi, Ollama Cloud, and QianwenAI. Providers without weekly data keep their existing presentation.

## Verification

- [x] `make test` — 1,721 tests, 3 skipped, 0 failures
- [x] `make build`
- [x] Ran the built macOS app and visually checked weekly display on/off, immediate updates, persistence, two independent providers, parent-switch fallback, and the existing menu
- [x] Rendered and inspected 0%, low, 50%, high, and 100% weekly states
- [x] Verified the ring adds zero horizontal width and remains a native template image for menu-bar tinting

## Notes

- Existing five-hour menu-bar eligibility is intentionally unchanged; enabling weekly display does not make hidden or weekly-only providers appear.
- Exact weekly reset information remains in the detailed menu to keep the top menu bar compact.


vinzdg:
Merged as `83a844e`. Thanks @ozgurasan-agent — the weekly ring is self-contained, off by default, and the catalog work is clean (order preserved, two new keys, no format specifiers to get wrong).

Note that this contained #299 in its entirety; I merged #299 first, so the duplicate content collapsed on merge. Worth avoiding in future — see my note on #301 — but nothing went wrong here.

One nit I left: the PR flips `Sources/Localizable.xcstrings` from no trailing newline to having one. Harmless, and arguably an improvement, so I've kept it rather than churn the file back.

PR #299, opened 2026-09-21
Reach the menu bar's own switch from the menu bar

ozgurasan-agent:
## Summary

Turning **Show limit information in menu bar** on or off meant opening Settings, finding Appearance, and scrolling to the bottom — for the one setting whose effect is two inches above, and whose whole point is to be glanced at rather than configured. Swapping between the icon and the readout is something people do on the day (the bar gets crowded, a talk starts, the numbers are wanted back), not once at install.

The menu the status item already opens now carries a checkable item for it, beside the other utilities.

**It is the one preference, not a second one.** The tick is drawn from `limits.isOn` — `Preferences.showsLimitsInMenuBar` mirrored into the controller by the sink that was already there — and clicking it writes that same `@Published` property. No new boolean, no new defaults key, no menu-local state, no second persistence path. Settings' switch and this one are two views of one value.

## Changes

- `StatusItemController` gains `onToggleLimits` and a checkable `NSMenuItem`, placed after a separator above `Refresh all` so it reads as a utility rather than as another provider row. A checkable item, not an embedded switch: the menu is a plain `NSMenu`, and a tick is how macOS puts a setting in one.
- `AppDelegate` wires that closure to `preferences.showsLimitsInMenuBar`.
- The limits sink moves from `RunLoop.main` to `DispatchQueue.main` — the same hazard the notch-size subscription below it already documents. Combine's RunLoop scheduler delivers in `.default` mode, which AppKit starves while it tracks an open menu, so a value switched from the menu's own tick would not land until the menu had closed.
- Reuses the existing catalog string, so the menu says what Settings says in all eight languages. No catalog change.

Nothing here is provider-specific: the toggle flips the master flag `StatusItemSummary.make` already honours for every provider.

## Test Plan

- [x] Run test suite (`make test-ci`) — **1714 tests, 0 failures**, 3 skipped. Four new tests in `Tests/StatusItemQuickToggleTests.swift`: the tick follows the preference; a click asks for the opposite; it writes the same preference and leaves providers, reset format and connections alone across off→on and a relaunch; it sits with the utilities without regressing the readings.
- [x] Tested on macOS (version: 26.0)
- [x] UI / Notch interactions verified

Verified in the running app on the real menu bar, not just by compiling:

- Preference OFF → menu opens with the item **unticked**; bar shows the default icon.
- Click → bar switches **instantly** to `✳ 84% · 2h 36m | ◎ 0% · 2h 37m`. One status item, no duplicate, no ghost menu, no crash — including the awkward case of enabling while the menu is open from the narrow default icon, where the item's width changes underneath it.
- With Settings open and visible, toggled OFF from the menu → the Settings switch flipped to OFF and its provider rows collapsed in the same instant, and the bar reverted to the default icon. No restart, no reopening Settings.
- Toggled ON again → Settings back ON with **Claude and Codex still selected**, both back on the bar.
- On disk afterwards: `showsLimitsInMenuBar = 1`, `menuBarProviders = (claude, codex)`, `resetTimeFormat = remaining` — providers and the other Menu Bar preferences untouched.
- `Refresh all`, `Settings…` and `Quit Codenotch` all still work; relaunched and the choice had survived, with the menu reopening ticked.

Menu with the preference on:

```
Claude — 84%
   Current session: 84% Used · 16% left · Resets Mon 13:00
   All models: 12% Used · 88% left · Resets Fri 19:00
Codex — 0%
   5h limit: 0% Used · 100% left · Resets Mon 13:01
   Weekly limit: 0% Used · 100% left · Resets Sep 28
──────────────────────────────────
✓ Show limit information in menu bar
──────────────────────────────────
Refresh all                      ⌘R
⚙ Settings...                     ⌘,
──────────────────────────────────
Quit Codenotch                   ⌘Q
```

Off, the same item is present and unticked, and the bar is the app's own icon.

## Note for the reviewer

One thing that cost me time and may cost you the same: the Settings switches render grey whenever that window is not the key window, which looks identical to OFF. When checking sync by eye, go by whether the Claude/Codex provider rows are present — they only show when the preference is actually on.


vinzdg:
Merged as `a8e1c62`. Thanks @ozgurasan-agent — 154 lines, one menu item, real tests, and no new strings (reusing the existing `"Show limit information in menu bar"` key is the right instinct). The `DispatchQueue.main` hop at `AppDelegate.swift:429-433` is correct and, better, documented: RunLoop default mode is starved while AppKit tracks a menu, which is the sort of thing that gets "simplified" away later without a comment.

This is the shape I'd like the rest of the menu-bar work to come in.

PR #301, opened 2026-09-21
Standardize provider usage in the status menu

ozgurasan-agent:
## Summary

- Standardize the detailed provider usage menu opened from the macOS status item.
- Present equivalent provider limits with shared labels, progress bars, thresholds, and reset copy.
- Keep zero-usage limits visible with an empty progress bar instead of dropping the row.
- Show relative and local absolute reset information from the same canonical timestamp.
- Keep the compact top menu bar reset-display behavior unchanged.

## Reset formatting

- Five-hour limits resetting on the same local calendar day show `(HH:mm)`.
- Five-hour limits crossing a calendar-day boundary show `(Day HH:mm)`.
- Weekly limits always show `(Day HH:mm)`.
- Calendar comparisons use Foundation's local calendar APIs and cover midnight, month/year boundaries, and daylight-saving transitions.

## Verification

- [x] `make test` — 1,738 tests passed; 3 opt-in live tests skipped.
- [x] `make run` — Debug build succeeded and launched.
- [x] Rendered the real SwiftUI provider section with five-hour and weekly fixtures and visually checked layout and clipping.


vinzdg:
@ozgurasan-agent — closing this one. I've merged #299 and #298; this needs to come back smaller, and I want to explain why rather than just decline it.

**It re-ships the other two.** `Sources/App/StatusItemSummary.swift` is byte-identical between this and #298, and `Tests/StatusItemQuickToggleTests.swift` appears verbatim in #299, #298 and here. #299 ⊂ #298 ⊂ #301 as literal supersets. That isn't three parallel proposals, it's one stack presented as three — and it means reviewing this meant reviewing the other two again.

**Two things in it reverse decisions rather than extend them:**

1. `Sources/Model/ResetCopy.swift:110-125` — `detailedLimitText` hardcodes `format: .remaining` and a fixed `"EEE HH:mm"`/`"HH:mm"`. It doesn't delete the `ResetTimeFormat` threading I added on Saturday, but it routes the new menu rows around it, so with **Reset date** selected the menu still says "Resets in 3h 20m". The fixed hour fields also contradict `ResetCopy.swift:70-78`, which uses `j` rather than `h` *because a pinned hour symbol was a shipped bug*. The comment at :121 calls the 24-hour choice intentional, which tells me that history wasn't visible to you — fair enough, but it has to hold.
2. `Sources/Settings/SettingsView.swift:1484,1504` moves the **Reset time** picker into a section titled "Menu Bar", and shows it even when `appPresence != .menuBar`. That's a global preference — `AppDelegate.swift:553` applies it to the notch fleet. Its own comment at :1502 concedes the sharing.

Also worth knowing for next time: `Tests/MenuBarSettingsSectionTests.swift:12-59` renders off-screen and asserts the bitmap has more than three distinct colours and that the values it just set are still set. A read-only render cannot fail either. And `Tests/AntigravityTests.swift:1345-1349` deletes real assertions on the menu's text ("Weekly limit", "29% Used · 71% left") and replaces them with `XCTAssertNotNil(menu.items[0].view)` — that's coverage removed, not moved.

**What I'd take, gladly:** the menu-rendering rewrite on its own, as an issue first so we can agree the shape before you build it; and the Settings reorganisation separately. One at a time, each a few hundred lines. You've built something people are using — #256 shipped in 1.16.0 and #298 and #299 just landed — and I'd rather keep that moving than have it stall in a 1,400-line review.

One standing request: if a PR adds a background monitor, say in the description what its idle wake-up rate is and what gates it. That's the thing I can't see from a diff and the thing most likely to get reverted later.

